PR #12, opened 2017-09-20
Add variable for custom radiusd startup params.

afoldesi-gds:
In order for us to be able to configure staging and live separately regards to detailed logging of the RADIUS traffic, we'll need the ability to add a parameter at the startup of the server.

See https://github.com/alphagov/govwifi-build/pull/7 for the private part of this change.


govwifi-frontend/variables.tf
```diff
@@ -67,3 +67,6 @@ variable "ithc-sg-list" {
   type    = "list"
   default = []
 }
+variable "radiusd-params" {
+  default = ""
+}
```
timmow: No newline at end of file - this breaks cat and leads to issues like we saw with adding the bastion key

timmow:
See also https://github.com/alphagov/govwifi-build/pull/7

govwifi-frontend/variables.tf
afoldesi-gds: Well, we should seldom edit this file with cat and generally speaking not spend an hour on the task of adding 3 lines on in 3 files but newlines should indeed be at the end of every file as a general rule :D good catch.

PR #25, opened 2018-03-26
Add new alarms for BurstBalance and ReplicaLag on RDS

RyanMacG:
We have encountered issues recently with the BurstBalance being heavily depleted
which in turn causes the Read Replica to lag behind. To highlight when
these situations occur we want to add better reporting around these metrics.

### DIFF
--- a/govwifi-backend/db.tf
+++ b/govwifi-backend/db.tf
@@ -213,6 +213,66 @@ resource "aws_cloudwatch_metric_alarm" "db_storagealarm" {
   treat_missing_data = "breaching"
 }
 
+resource "aws_cloudwatch_metric_alarm" "db_burstbalancealarm" {
+  count               = "${var.db-instance-count}"
+  alarm_name          = "${var.Env-Name}-db-burstbalanace-alarm"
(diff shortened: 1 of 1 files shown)


govwifi-backend/db.tf
```diff
@@ -213,6 +213,66 @@ resource "aws_cloudwatch_metric_alarm" "db_storagealarm" {
...
+  namespace           = "AWS/RDS"
+  period              = "60"
+  statistic           = "Minimum"
+  threshold           = "300"
+
+  dimensions {
+    DBInstanceIdentifier = "${aws_db_instance.read_replica.identifier}"
+  }
+
+  alarm_description  = "This metric monitors the IOPS burst balance available for the DB read replica."
```
adrianclay: This description talks about IOPS?

RyanMacG: Don't understand the question/comment

adrianclay: GitHub rendered the diff poorly, you've already fixed the one it should be showing again.

PR #26, opened 2018-03-28
Performance testing module setup

afoldesi-gds:
Creates a new module for setting up an instance for running performance testing within a specified subnet. Most parameters of the instance are configurable. 

Also contains fixes for the issues uncovered during the setup of the new environment.
Namely fixes for overcoming ELB SSL cert catch-22 and bug with elasticache id length.

### DIFF
--- a/govwifi-backend/cache.tf
+++ b/govwifi-backend/cache.tf
@@ -5,7 +5,7 @@ resource "aws_elasticache_subnet_group" "cache-subnets" {
 }
 
 resource "aws_elasticache_cluster" "cache" {
-  cluster_id             = "cache-${var.Env-Name}-wifi"
+  cluster_id             = "cache-${length(var.Env-Name)}-wifi"
   engine                 = "memcached"
   node_type              = "${var.cache-node-type}"
   port                   = 11211
--- a/govwifi-backend/cache.tf
+++ b/govwifi-backend/cache.tf
@@ -5,7 +5,7 @@ resource "aws_elasticache_subnet_group" "cache-subnets" {
 }
 
 resource "aws_elasticache_cluster" "cache" {
-  cluster_id             = "cache-${length(var.Env-Name)}-wifi"
+  cluster_id             = "${var.Env-Name}-wifi"
   engine                 = "memcached"
   node_type              = "${var.cache-node-type}"
   port                   = 11211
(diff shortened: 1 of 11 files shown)


govwifi-backend/cache.tf
```diff
@@ -5,7 +5,7 @@ resource "aws_elasticache_subnet_group" "cache-subnets" {
 }
 
 resource "aws_elasticache_cluster" "cache" {
-  cluster_id             = "cache-${var.Env-Name}-wifi"
+  cluster_id             = "cache-${length(var.Env-Name)}-wifi"
```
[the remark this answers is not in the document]
afoldesi-gds: Sure it would. There's also the more pressing issue if this not working at all as the word "performance" is apparently too long. (limit being 20) 

We could change how we construct the cluster id for all environments. Quite frankly we know it's a cache and in which region it is in - we might as well just use the product and environment name.

govwifi-backend/cache.tf
```diff
@@ -5,7 +5,7 @@ resource "aws_elasticache_subnet_group" "cache-subnets" {
 }
 
 resource "aws_elasticache_cluster" "cache" {
-  cluster_id             = "cache-${var.Env-Name}-wifi"
+  cluster_id             = "${var.Env-Name}-wifi"
```
afoldesi-gds: When this update is released to live the cache will be flushed which marries up well with the expiry changes here: https://github.com/alphagov/govwifi/pull/120

govwifi-backend/modules/ecs-autoscaling/ecs-autoscaling.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_autoscaling_group" "ecs-cluster" {
-  availability_zones        = ["${split(",", var.availability_zones)}"]
+#  availability_zones        = ["${split(",", var.availability_zones)}"]
```
afoldesi-gds: This was the horrible change without which the infrastructure could not possibly be built by the existing terraform code and the issue had to be reported to Milan.

govwifi-backend/cache.tf
afoldesi-gds: To avoid character limit on the cliuster_id, we need to remove the "cache-" prefix.

afoldesi-gds:
Ok. So this should have been merged. No probs tho. It'll be fine.

PR #27, opened 2018-03-29
Increase period and no longer breach on missing data for BurstBalance

RyanMacG:
After putting the new BurstBalance alarams in place we found that the
data can sometimes take several minutes to appear in cloudwatch causing
it to think the data is missing and instantly breach even though the
Balance is still 100%

### DIFF
--- a/govwifi-backend/db.tf
+++ b/govwifi-backend/db.tf
@@ -220,7 +220,7 @@ resource "aws_cloudwatch_metric_alarm" "db_burstbalancealarm" {
   evaluation_periods  = "1"
   metric_name         = "BurstBalance"
   namespace           = "AWS/RDS"
-  period              = "60"
+  period              = "180"
   statistic           = "Minimum"
   threshold           = "45"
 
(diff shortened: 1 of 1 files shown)


govwifi-backend/db.tf
```diff
@@ -220,7 +220,7 @@ resource "aws_cloudwatch_metric_alarm" "db_burstbalancealarm" {
   evaluation_periods  = "1"
   metric_name         = "BurstBalance"
   namespace           = "AWS/RDS"
-  period              = "60"
+  period              = "180"
```
adrianclay: How did you come to this figure?

RyanMacG: We decided one minute was too short a period when we discussed it and 3 minutes felt like a reasonable alarm point; if the balance is sub 45% for 3 minutes we can start being concerned?

RyanMacG: No, I've been unable to find anything indicating why this would be happening

adrianclay: We should monitor if this continues to be an issue then and increase the period if necessary.

PR #30, opened 2018-04-10
Update retention policy for cloudwatch

RyanMacG:
As part of our GDPR compliance work we only want to store los for 90
days.


govwifi-frontend/cluster.tf
```diff
@@ -6,8 +6,9 @@ resource "aws_ecs_cluster" "frontend-cluster" {
 
 resource "aws_cloudwatch_log_group" "frontend-log-group" {
   name = "${var.Env-Name}-frontend-docker-log-group"
+
   #keep the logs forever
```
adrianclay: Comment is misleading now.

PR #33, opened 2018-04-25
Add new bucket to store Performance Platform data backups.

afoldesi-gds:
Specific policy is added to the bastion host, so it can write to the new bucket.


afoldesi-gds:
Ping.

PR #34, opened 2018-04-25
Add lambda module

RyanMacG:
This allows us to deploy two lambda instances that can be triggered with
cron cloudwatch events to delete user and session data

### DIFF
--- a/govwifi-lambda/cloudwatch.tf
+++ b/govwifi-lambda/cloudwatch.tf
@@ -1,0 +1,39 @@
+resource "aws_cloudwatch_event_rule" "every_day_at_midnight" {
+  name                = "every-day-at-midnight"
+  description         = "Triggers every morning at Midnight"
+  schedule_expression = "cron(0 0 * * ? *)"
+}
+
+resource "aws_cloudwatch_event_target" "delete_users_every_day_at_midnight" {
+  rule      = "${aws_cloudwatch_event_rule.every_day_at_midnight.name}"
+  target_id = "user_deletion"
+  arn       = "${aws_lambda_function.user_deletion.arn}"
+}
+
+resource "aws_lambda_permission" "allow_cloudwatch_to_call_delete_users" {
+  statement_id  = "AllowExecutionFromCloudWatch"
+  action        = "lambda:InvokeFunction"
+  function_name = "${aws_lambda_function.user_deletion.function_name}"
+  principal     = "events.amazonaws.com"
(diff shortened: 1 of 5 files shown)


govwifi-lambda/cloudwatch.tf
```diff
@@ -0,0 +1,39 @@
+resource "aws_cloudwatch_event_rule" "every_morning_at_one" {
+  name                = "every-morning-at-one"
+  description         = "Triggers every morning at 1AM"
+  schedule_expression = "cron(0 1 * * ? *)"
```
afoldesi-gds: One o'clock is probably a bit busy - that's when the flat file dump is running currently, hogging some res. How about 0:37?  or 42? :) - depending how long we expect this to run, maybe even sooner. 

The reason for not running it after the dump is so the flat file backups and the AWS backups running between 3 and 4 are in sync.

govwifi-lambda/cloudwatch.tf
```diff
@@ -0,0 +1,39 @@
...
+  action        = "lambda:InvokeFunction"
+  function_name = "${aws_lambda_function.user_deletion.function_name}"
+  principal     = "events.amazonaws.com"
+  source_arn    = "${aws_cloudwatch_event_rule.every_morning_at_one.arn}"
+}
+
+resource "aws_cloudwatch_event_rule" "every_morning_at_two" {
+  name                = "every-morning-at-two"
+  description         = "Triggers every morning at 2AM"
+  schedule_expression = "cron(0 2 * * ? *)"
```
afoldesi-gds: See above. We most likely either want to run these between midnight and one or move the flat file dump to 2 or 2:30.

govwifi-lambda/cloudwatch.tf
```diff
@@ -0,0 +1,39 @@
+resource "aws_cloudwatch_event_rule" "every_morning_at_one" {
+  name                = "every-morning-at-one"
+  description         = "Triggers every morning at 1AM"
+  schedule_expression = "cron(0 1 * * ? *)"
+}
+
+resource "aws_cloudwatch_event_target" "delete_users_every_morning_at_one" {
+  rule      = "${aws_cloudwatch_event_rule.every_morning_at_one.name}"
+  target_id = "user_deletion"
```
afoldesi-gds: We probably should talk about soft deletion of users and sponsors.

adrianclay:
This links up to the variables here https://github.com/alphagov/govwifi-build/pull/50/files#diff-e29669a4a7d32f1a7ba74b37cc6e893bR230

PR #40, opened 2018-05-10
Allowed Sites API ECS Cluster and Infrastructure

emileswarts:
Create a new ECS backend cluster for the Ruby backend


govwifi-backend-ruby/modules/ecs-autoscaling/launch-configuration.tf
```diff
@@ -0,0 +1,225 @@
+resource "aws_launch_configuration" "ecs" {
+  name_prefix          = "tf-${var.Env-Name}-ruby-lc-"
+  image_id             = "${var.ami}"
+  instance_type        = "${var.instance_type}"
+  iam_instance_profile = "${var.instance-profile-id}"
+  key_name             = "${var.key_name}"
+  security_groups      = ["${split(",", var.security_group_ids)}"]
+
+  user_data = <<DATA
```
adrianclay: Is this a dupe of the user_data from the other launch-config?

Be good to remove that duplication.

govwifi-backend-ruby/cluster.tf
```diff
@@ -0,0 +1,123 @@
...
+      "disableNetworking": null,
+      "dnsSearchDomains": null,
+      "portMappings": [
+        {
+          "hostPort": 80,
+          "containerPort": 80,
+          "protocol": "tcp"
+        },
+        {
+          "hostPort": 8080,
```
adrianclay: 8080 ?

govwifi-backend-ruby/modules/ecs-autoscaling/launch-configuration.tf
emileswarts: We tried to solve this a while ago, by extracting it into a reusable file we could read.  
It was more complex than we anticipated.  
Propose we allow this to exist for now and add a ticket to come back to it.

Introducing Ansible for provisioning would also address this.

govwifi-allowed-sites-api/loadbalancer.tf
```diff
@@ -0,0 +1,53 @@
...
+    lb_port           = 8080
+    lb_protocol       = "http"
+  }
+
+  listener {
+    instance_port      = 80
+    instance_protocol  = "http"
+    lb_port            = 443
+    lb_protocol        = "https"
+    ssl_certificate_id = "${var.elb-ssl-cert-arn}"
```
afoldesi-gds: Are these re-generated for allowed-sites-api-elb....gov.uk?

govwifi-allowed-sites-api/loadbalancer.tf
```diff
@@ -0,0 +1,53 @@
+resource "aws_elb" "allowed-sites-api-elb" {
+  count           = "${var.backend-elb-count}"
+  name            = "allowed-sites-api-elb-${var.Env-Name}"
+  subnets         = ["${var.subnet-ids}"]
+  security_groups = ["${var.elb-sg-list}"]
+
+  listener {
```
afoldesi-gds: Could probably get rid of the HTTP listeners here if we won't use them (the HTTPS listener rules below will "translate" requests to HTTP on the instance)

govwifi-allowed-sites-api/loadbalancer.tf
emileswarts: Yeah just generated one before running the terraform

PR #41, opened 2018-05-10
Use unique directory name for temp files.

afoldesi-gds:
Found it.

This fixes the issue with the tmp directory we were seeing on the bastion hosts.


afoldesi-gds:
Probably not. That command actually uses the /tmp directory, so should it get removed again, it will fail.
Eg.
`
:~# mktemp -d
mktemp: failed to create directory via template ‘/tmp/tmp.XXXXXXXXXX’: No such file or directory
`

PR #44, opened 2018-05-14
Forward to 8080 on the instances from the ELB

emileswarts:


afoldesi-gds:
You will need to listen on 80 as well for the health checks tho

PR #54, opened 2018-06-05
Authentication api

emileswarts:
Create an ALB with a target group which scales containers on ec2 instances.
This is the new infrastructure for the Authentication API


govwifi-api/cluster.tf
```diff
@@ -0,0 +1,93 @@
...
+}
+
+resource "aws_ecs_task_definition" "authorisation-api-task" {
+  family = "authorisation-api-task-${var.Env-Name}"
+
+  container_definitions = <<EOF
+[
+    {
+      "volumesFrom": [],
+      "memory": 1900,
```
adrianclay: Where did this number come from?

RyanMacG: I'd assume from the existing clusters

RyanMacG: @adrianclay https://github.com/alphagov/govwifi-terraform/blob/master/govwifi-backend/cluster.tf#L20 lower than this limit

govwifi-api/cluster.tf
```diff
@@ -0,0 +1,93 @@
...
+          "name": "DB_USER",
+          "value": "${var.db-user}"
+        },{
+          "name": "DB_HOSTNAME",
+          "value": "db.${lower(var.aws-region-name)}.${var.Env-Subdomain}.service.gov.uk"
+        },{
+          "name": "ENVIRONMENT_NAME",
+          "value": "${var.Env-Name}"
+        },{
+          "name": "FRONTEND_API_KEY",
```
adrianclay: Does this get used?

adrianclay:
Won't we require a `aws_appautoscaling_target` ?

https://www.terraform.io/docs/providers/aws/r/appautoscaling_target.html#ecs-service-autoscaling

govwifi-api/elb.tf
```diff
@@ -0,0 +1,60 @@
...
+  load_balancer_type = "application"
+
+  tags {
+    Name = "api-alb-${var.Env-Name}"
+  }
+}
+
+resource "aws_alb_listener" "alb_listener" {
+  load_balancer_arn = "${aws_lb.api-alb.arn}"
+  port              = "8080"
```
adrianclay: Doesn't 443 makes more sense for HTTPS?

govwifi-api/scaling-policy.tf
```diff
@@ -0,0 +1,27 @@
...
+resource "aws_cloudwatch_metric_alarm" "cpualarm" {
+  count               = "${var.backend-cpualarm-count}"
+  alarm_name          = "${var.Env-Name}-api-cpu-alarm"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "2"
+  metric_name         = "CPUUtilization"
+  namespace           = "AWS/EC2"
+  period              = "60"
+  statistic           = "Average"
+  threshold           = "50"
```
adrianclay: Be good to see this alarm + autoscaling tested and demonstrated as working on staging.

govwifi-api/cluster.tf
emileswarts: This needs to be looked at when we performance test,  t2.medium has 4G ram, this is saying take just under half of it for the task.

PR #58, opened 2018-06-07
Update to re-enable ECS agent update

RyanMacG:
We've been receiving warnings that the ECS agents are out of date on
containers, we've confirmed the upgrade works with the authentication
containers so are rolling this out to the remaining containers.


adrianclay:
When are you planning to apply these changes?

PR #60, opened 2018-06-11
Autoscaling auth

emileswarts:
Allow our Authentication application to scale to meet the current demand.
Both EC2 and ECS scale up and down using autoscaling policies.

### DIFF
--- a/govwifi-api/scaling-policy.tf
+++ b/govwifi-api/scaling-policy.tf
@@ -25,3 +25,91 @@ resource "aws_cloudwatch_metric_alarm" "cpualarm" {
   alarm_actions      = ["${aws_autoscaling_policy.scale-policy.arn}", "${var.critical-notifications-arn}"]
   treat_missing_data = "breaching"
 }
+
+resource "aws_cloudwatch_metric_alarm" "auth_service_high" {
+  alarm_name          = "${var.Env-Name}-api-cpu-alarm-high-30"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "1"
+  metric_name         = "CPUUtilization"
+  namespace           = "AWS/ECS"
+  period              = "60"
+  statistic           = "Average"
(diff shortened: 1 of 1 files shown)


govwifi-api/scaling-policy.tf
```diff
@@ -25,3 +25,91 @@ resource "aws_cloudwatch_metric_alarm" "cpualarm" {
...
+  resource_id        = "service/${aws_ecs_cluster.api-cluster.name}/${aws_ecs_service.authorisation-api-service.name}"
+  policy_type        = "StepScaling"
+  scalable_dimension = "ecs:service:DesiredCount"
+
+  step_scaling_policy_configuration {
+    adjustment_type    = "ChangeInCapacity"
+    metric_aggregation_type   = "Average"
+
+    step_adjustment {
+      metric_interval_upper_bound = 0
```
RyanMacG: I'd add a cooldown here of maybe 2/3 minutes. Looks like the default is 5 so if it hits the cap and then settles down we'll have extra instances hanging around for about 20 minutes

govwifi-api/scaling-policy.tf
```diff
@@ -25,3 +25,91 @@ resource "aws_cloudwatch_metric_alarm" "cpualarm" {
...
+  name               = "ECS Scale Down"
+  service_namespace  = "ecs"
+  resource_id        = "service/${aws_ecs_cluster.api-cluster.name}/${aws_ecs_service.authorisation-api-service.name}"
+  policy_type        = "StepScaling"
+  scalable_dimension = "ecs:service:DesiredCount"
+
+  step_scaling_policy_configuration {
+    adjustment_type    = "ChangeInCapacity"
+    metric_aggregation_type   = "Average"
+
```
RyanMacG: I'd add a cooldown here of maybe 2/3 minutes. Looks like the default is 5 so if it hits the cap and then settles down we'll have extra instances hanging around for about 20 minutes

PR #65, opened 2018-06-12
Change ECS placement to be spread out over instances

emileswarts:
Previously we would see two tasks running on one instance with another
instance in the fleet having no tasks running.

By specifying this placement strategy, it will place containers based on
cpu load.

### DIFF
--- a/govwifi-api/cluster.tf
+++ b/govwifi-api/cluster.tf
@@ -88,7 +88,12 @@ resource "aws_ecs_service" "authorisation-api-service" {
   desired_count   = "${var.backend-instance-count}"
   iam_role        = "${var.ecs-service-role}"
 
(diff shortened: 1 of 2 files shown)


govwifi-api/cluster.tf
```diff
@@ -88,6 +88,11 @@ resource "aws_ecs_service" "authorisation-api-service" {
   desired_count   = "${var.backend-instance-count}"
   iam_role        = "${var.ecs-service-role}"
 
+  placement_strategy {
+    type  = "binpack"
```
RyanMacG: shouldn't this be spread and using the availabilty zones?

RyanMacG:
Depends on https://github.com/alphagov/govwifi-build/pull/106

PR #77, opened 2018-06-22
Add S3 bucket for storing the wordlist

RyanMacG:
The wordlist gets used during signup to generate a random password for the user.  We don't want to expose the wordlist in GitHub, so to avoid this we'll push it into S3 and give permission to Jenkins to pull it down while building the image.

### DIFF
--- a/govwifi-api/wordlist.tf
+++ b/govwifi-api/wordlist.tf
@@ -1,0 +1,42 @@
+resource "aws_s3_bucket" "wordlist" {
+  bucket = "govwifi-wordlist"
+  count  = "${var.wordlist-bucket-count}"
+  acl    = "private"
+
+  tags {
+    Name = "wordlist-bucket"
+  }
+}
+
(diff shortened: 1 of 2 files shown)


govwifi-api/wordlist.tf
```diff
@@ -0,0 +1,42 @@
...
+resource "aws_s3_bucket_object" "wordlist" {
+  bucket = "${aws_s3_bucket.wordlist.bucket}"
+  count  = "${var.wordlist-bucket-count}"
+  key    = "wordlist-short"
+  source = "${var.wordlist-file-path}"
+  etag   = "${md5(file(var.wordlist-file-path))}"
+}
+
+resource "aws_iam_user_policy" "jenkins-read-wordlist-policy" {
+  user = "${aws_iam_user.jenkins-read-wordlist-bucket.name}"
```
chrisblackburn: The inconsistent whitespace between the declaration and the equals upsets me. Would prefer one or t'other

PR #80, opened 2018-06-28
Introduce a new variable name for the read replica database

adrianclay:
Previously we were overwriting the db hostname with a read replica
value for production.  This was causing trouble for user signups
as we need to insert data.  Inserting data into a read replica
blows up when running that on a read replica.

By separating out the two variables we can specify the read replica
for the authroisations task, and the master for the user signup
task.

### DIFF
--- a/govwifi-api/variables.tf
+++ b/govwifi-api/variables.tf
@@ -32,6 +32,8 @@ variable "db-password" {}
 
 variable "db-hostname" {}
 
+variable "db-read-replica-hostname" {}
+
 variable "rack-env" {}
 
 variable "radius-server-ips" {}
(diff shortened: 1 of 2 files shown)


govwifi-api/variables.tf
```diff
@@ -32,6 +32,8 @@ variable "db-password" {}
 
 variable "db-hostname" {}
 
+variable "read-replica-db-hostname" {}
```
chrisblackburn: See question on whether `db-` is a prefix convention that should be followed. LMK thoughts and can approve either way

adrianclay: I think it's a fair enough comment.  Have updated.

PR #81, opened 2018-07-02
Move Manually Created SES Receipt Rules into Terraform

emileswarts:
We want this to be managed by Terraform as it involves less manual
effort moving forward.

Creating the rule set itself is still manual as this causes errors with
Terraform applying this accross multiple environments.

Delete the unused test rules which belonged to an inactive test rule set.
Recreate the rules that were manually created.

This doesn't make the new rule set active, it only creates a copy which
can be made active once we are happy.

### DIFF
--- a/govwifi-emails/emails.tf
+++ b/govwifi-emails/emails.tf
@@ -1,18 +1,11 @@
-# SES setup to receive emails - there should be a DNS MX entry for
-# the amazon mail server in the region used.
-resource "aws_ses_receipt_rule_set" "main-ses-ruleset" {
-  rule_set_name = "${var.Env-Name}-ses-ruleset"
-}
-
-resource "aws_ses_receipt_rule" "incoming-ses-rule" {
-  name          = "${var.Env-Name}-incoming-ses-rule"
-  rule_set_name = "${var.Env-Name}-ses-ruleset"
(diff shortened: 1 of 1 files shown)


govwifi-emails/emails.tf
```diff
@@ -1,28 +1,53 @@
...
 
   depends_on = [
-    "aws_ses_receipt_rule_set.main-ses-ruleset",
     "aws_sns_topic.govwifi-email-notifications",
     "aws_s3_bucket.emailbucket",
+    "aws_ses_receipt_rule.user-signup-rule",
   ]
 
   recipients = [
     "enrol@${var.Env-Subdomain}.service.gov.uk",
```
adrianclay: We can remove enrol, enroll and signup from this rule now.  Given they are being processed by the rule above.

emileswarts: Yeah noticed that but wanted to preserve current functionality and address that in a future PR.
Just to minimise risk on activating this new rule set for now.

emileswarts:
Have opened an issue regarding this SES Rule Set being a manual step.

govwifi-emails/emails.tf
adrianclay: Fair enough.

PR #88, opened 2018-07-12
Remove notify user signup email template id environment variable

emileswarts:
This isn't needed anymore because we store this config locally in the
user signup API application


adrianclay:
Links up to https://github.com/alphagov/govwifi-user-signup-api/pull/37 for future reference.

PR #90, opened 2018-07-16
Basic infrastructure for the new Logging API

emileswarts:
This API will deal with post-auth and accounting.
We have a sinatra application ready and this creates the infrastructure
for it.

depends on https://github.com/alphagov/govwifi-build/pull/130

### DIFF
--- a/govwifi-api/logging-api-cluster.tf
+++ b/govwifi-api/logging-api-cluster.tf
@@ -1,0 +1,142 @@
+resource "aws_cloudwatch_log_group" "logging-api-log-group" {
+  name = "${var.Env-Name}-logging-api-docker-log-group"
+
+  retention_in_days = 90
+}
+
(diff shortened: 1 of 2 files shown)


govwifi-api/logging-api-cluster.tf
```diff
@@ -0,0 +1,142 @@
...
+}
+
+resource "aws_ecs_task_definition" "logging-api-task" {
+  family   = "logging-api-task-${var.Env-Name}"
+
+  container_definitions = <<EOF
+[
+    {
+      "volumesFrom": [],
+      "memory": 1900,
```
adrianclay: Are we going to run out of memory on our current cluster nodes, given we've given each container loads of memory?

emileswarts: We give each container just less than 50% of the instance memory.
We will require a new instance to run 2 tasks of the logging API.  This is already happening on staging.

PR #93, opened 2018-07-19
Disable the survey sending behaviour

adrianclay:
We're currently not reviewing the surveys received back, instead the focus of the team is feedback from organisations that are installing GovWifi into their buildings.

Additionally we have the goal of re-writing the application from PHP to Ruby.  The surveys code being something that is currently written in PHP which would need to be rewritten if we want to keep the
functionality.

In the interests of minimising waste, we have decided to not rewrite the survey behaviour now, and instead re-implement it when we want to iterate on the end-user experience.  At which point we'd have
capacity inside the team to review, and categorise feedback.

### DIFF
--- a/govwifi-backend/management.tf
+++ b/govwifi-backend/management.tf
@@ -102,11 +102,6 @@ Content-Type: text/x-shellscript; charset="us-ascii"
 #!/bin/bash
 # Set up cron scripts for timed jobs
 
-cat <<'EOF' > ./ping-survey
-#!/bin/bash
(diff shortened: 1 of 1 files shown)


adrianclay:
I propose we merge this alongside disabling the PHP performance platform crons, as it is very similar functionality and should be ready Monday.  This will reduce the amount of duplicate deploy work.

naliwajek:
Sounds good to me. Though, if you think about all 5 performance metrics, I'd say Tuesday seems more feasible.

PR #96, opened 2018-07-23
Add ENV vars for Performance platform

naliwajek:


govwifi-api/user-signup-api-cluster.tf
```diff
@@ -105,6 +105,18 @@ resource "aws_ecs_task_definition" "user-signup-api-task" {
         },{
```
naliwajek: `user-signup-api-cluster` sounded like something that should correspond to `user-signup-api` repo. Let me know if my assumption is wrong.

PR #97, opened 2018-07-23
Add Autoscaling to Logging API

emileswarts:
This adds 2 Cloudwatch alarms, one when the CPU is above 50% for 5
minutes and another when the CPU is below 10% for 10 minutes.

These alarms will trigger an autoscaling action, adding or removing
another task.

### DIFF
--- a/govwifi-api/alarms-logging.tf
+++ b/govwifi-api/alarms-logging.tf
@@ -1,0 +1,45 @@
+resource "aws_cloudwatch_metric_alarm" "logging-ecs-cpu-alarm-high" {
+  alarm_name          = "${var.Env-Name}-logging-ecs-cpu-alarm-high"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "1"
+  metric_name         = "CPUUtilization"
+  namespace           = "AWS/ECS"
+  period              = "300"
+  statistic           = "Average"
+  threshold           = "50"
+
+  dimensions {
+    ClusterName = "${aws_ecs_cluster.api-cluster.name}"
+    ServiceName = "${aws_ecs_service.logging-api-service.name}"
+  }
+
+  alarm_description  = "This alarm tells ECS to scale up based on high CPU - Logging"
+  alarm_actions      = [
+    "${aws_appautoscaling_policy.ecs-policy-up-logging.arn}",
+    "${var.critical-notifications-arn}"
+  ]
+
+  treat_missing_data = "breaching"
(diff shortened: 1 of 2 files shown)


govwifi-api/logging-scaling-policy.tf
```diff
@@ -0,0 +1,48 @@
+resource "aws_appautoscaling_target" "logging-ecs-target" {
+  service_namespace  = "ecs"
+  resource_id        = "service/${aws_ecs_cluster.api-cluster.name}/${aws_ecs_service.logging-api-service.name}"
+  max_capacity       = 20
```
kjdchapman: Should we have another alert for when we're at max capacity and the CPU usage is 50%+?

govwifi-api/alarms-logging.tf
```diff
@@ -0,0 +1,45 @@
...
+    "${var.critical-notifications-arn}"
+  ]
+
+  treat_missing_data = "breaching"
+}
+
+resource "aws_cloudwatch_metric_alarm" "logging-ecs-cpu-alarm-low" {
+  alarm_name          = "${var.Env-Name}-logging-ecs-cpu-alarm-low"
+  comparison_operator = "LessThanThreshold"
+  evaluation_periods  = "2"
```
kjdchapman: Nitpicking, but the `low` alarm is 2 and the `high` alarm is 1 - is that on purpose?

kjdchapman: (evaluation periods, that is)

govwifi-api/logging-scaling-policy.tf
emileswarts: If we ever do reach 20 tasks and it is unable to scale any more, we will start getting health check failures.  The service will need to get hammered to even trigger 1 autoscaling action.  Current CPU is around 2% for context.

govwifi-api/alarms-logging.tf
emileswarts: It is on purpose yeah, scaled down is allowed to take twice as long to trigger an event because cleaning up is not mission critical.

govwifi-api/logging-scaling-policy.tf
kjdchapman: Ah cool, forgot about the healthcheck failures :+1:

govwifi-api/alarms-logging.tf
kjdchapman: Makes sense!

PR #103, opened 2018-07-31
Add infrastructure for hosting the admin platform

adrianclay:
We're placing this into a new EC2 container cluster as it will be a public application vs the API cluster which is private.

Any DOS attacks to the public admin platform shouldn't affect the rest of the service.

### DIFF
--- a/govwifi-admin/acm.tf
+++ b/govwifi-admin/acm.tf
@@ -1,0 +1,17 @@
+resource "aws_acm_certificate" "admin_platform_cert" {
+  domain_name = "${aws_route53_record.admin_platform.name}"
+  validation_method = "DNS"
+}
+
+resource "aws_route53_record" "admin_platform_cert_validation" {
+  name = "${aws_acm_certificate.admin_platform_cert.domain_validation_options.0.resource_record_name}"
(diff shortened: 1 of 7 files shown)


adrianclay:
I'm a little disappointed about the amount of duplication we have between the different modules.  My thought for removing this duplication would be to introduce a module that our other modules could use, but I'm concerned about refactoring production infrastructure wanting to tear down what has been created and re-creating it.  I wonder if anybody has had any experience with this?

emileswarts:
Your suggestion of extracting a module that other modules use is the only thing I can think of.
The userdata would be a prime candidate because it's so verbose.

Would add it to the backlog, if we do address it, it would be good to do it as an isolated task.

adrianclay:
My fear with extracting it as a module will be that the resource ids will change, and so terraform will want to recreate the resources unless we do something like is explained here https://www.terraform.io/docs/commands/state/mv.html#example-move-a-resource-into-a-module

adrianclay:
Also the RE jenkins terraform is an example of how you can do nice cloud inits https://github.com/alphagov/re-build-systems/tree/master/terraform/jenkins/cloud-init

PR #104, opened 2018-08-03
Disable cron for sending Performance Platform statistics

emileswarts:
We have rewritten the PHP performance platform statistics functionality
in Ruby.  This uses ECS scheduled tasks as opposed to traditional cron.

Disable the cron here that uses the PHP backend for sending performance
statistics.  This is now the responsibility of the Ruby applications.

### DIFF
--- a/govwifi-backend/management.tf
+++ b/govwifi-backend/management.tf
@@ -102,16 +102,6 @@ Content-Type: text/x-shellscript; charset="us-ascii"
 #!/bin/bash
 # Set up cron scripts for timed jobs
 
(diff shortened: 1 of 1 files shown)


kjdchapman:
Out of interest, what does the remaining `/backup-performanceplatform` do?

emileswarts:
@kjdchapman You can see what it does here:
https://github.com/alphagov/govwifi-terraform/blob/9843f845d84ae53c8dbf38e7e87783175c898fac/govwifi-backend/management.tf#L105

PR #117, opened 2018-08-16
Expose port 3000 of Ruby healthcheck app in ECS

naliwajek:


naliwajek:
Related: https://github.com/alphagov/govwifi-build/pull/168

PR #122, opened 2018-09-03
Set Sentry environment variables

emileswarts:
The task definition will contain the Sentry DSN to use in the admin


kjdchapman:
Can't formally review, but :+1: from me

PR #132, opened 2018-09-20
Create S3 bucket policy for Radius whitelist config

emileswarts:
We will be getting the radius whitelist configuration from S3 instead of
the allowed sites API.

This allows the frontends to get this configuration from bucket.


adrianclay:
Why did you decide to whitelist via the IP addresses instead of a role for the frontend EC2s?

PR #133, opened 2018-09-20
Populate the S3 object key for the whitelist

emileswarts:
Our admin platform will populate this configuration file.
Eventually the Radius servers will read this file to populate the IP
whitelist.

### DIFF
--- a/govwifi-admin/cluster.tf
+++ b/govwifi-admin/cluster.tf
@@ -88,6 +88,9 @@ resource "aws_ecs_task_definition" "admin-task" {
         },{
           "name": "S3_PUBLISHED_LOCATIONS_IPS_OBJECT_KEY",
           "value": "ips-and-locations.json"
+        },{
+          "name": "S3_WHITELIST_OBJECT_KEY",
+          "value": "clients.conf"
         }
       ],
       "links": null,


govwifi-admin/cluster.tf
```diff
@@ -88,6 +88,9 @@ resource "aws_ecs_task_definition" "admin-task" {
         },{
           "name": "S3_PUBLISHED_LOCATIONS_IPS_OBJECT_KEY",
           "value": "ips-and-locations.json"
+        },{
+          "name": "S3_WHITELIST_OBJECT_KEY",
```
kjdchapman: Are both of the objects going in the same bucket? I don't know if I'm reading this change right.

emileswarts: Yes the bucket was named govwifi-admin-staging or govwifi-admin-production

tommotaylor: Same bucket different objects. Like saying "same folder, different files"

tommotaylor:
Looks good to me

govwifi-admin/cluster.tf
adrianclay: Be nice if there were one bucket in London for the London frontends and one bucket in Dublin for the Dublin frontends.

emileswarts: S3 Buckets are global.

emileswarts: I stand corrected, ignore that

PR #136, opened 2018-09-25
Create Authorisation API cluster in Ireland

naliwajek:
Currently the authentication backend only runs in London.
Spin up another API cluster in Ireland to make it more robust.

This adds flags for infrastructure we want to supress in Ireland
ie. Logging and User Signup


govwifi-api/user-signup-api-cluster.tf
```diff
@@ -1,15 +1,17 @@
 resource "aws_cloudwatch_log_group" "user-signup-api-log-group" {
+  count = "${var.user-signup-docker-image == ""  ? 0 : 1}"
```
emileswarts: Why not just introduce the count, and have it set to 1 in london and 0 in Ireland?

kjdchapman:
LGTM

PR #146, opened 2018-10-15
Add S3 bucket for MOUs in Admin

naliwajek:


govwifi-admin/s3.tf
```diff
@@ -15,6 +15,23 @@ resource "aws_s3_bucket" "admin-bucket" {
...
+  acl           = "private"
+
+  tags {
+    Name        = "${title(var.Env-Name)} MOUs from Admin"
+    Region      = "${title(var.aws-region-name)}"
+    Environment = "${title(var.rack-env)}"
+  }
+
+  versioning {
+    enabled = true
```
naliwajek: Not sure if we want to be that thorough with MOUs 🤔

PR #147, opened 2018-10-16
Add seperate S3 policy for MOU bucket

naliwajek:
ActiveStorage, which we use to interact with MOU bucket from within Rails, required a lot of permissions – so we separate permissions for those two buckets.

### DIFF
--- a/govwifi-admin/iam-roles.tf
+++ b/govwifi-admin/iam-roles.tf
@@ -57,10 +57,17 @@ resource "aws_iam_role_policy" "ecs-admin-instance-policy" {
       "Action": [
         "s3:PutObject"
       ],
-      "Resource": [
-        "${aws_s3_bucket.admin-bucket.arn}/*",
-        "${aws_s3_bucket.admin-mou-bucket.arn}/*"
-      ]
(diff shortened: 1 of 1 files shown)


govwifi-admin/iam-roles.tf
```diff
@@ -57,10 +57,17 @@ resource "aws_iam_role_policy" "ecs-admin-instance-policy" {
...
         "s3:PutObject"
       ],
-      "Resource": [
-        "${aws_s3_bucket.admin-bucket.arn}/*",
-        "${aws_s3_bucket.admin-mou-bucket.arn}/*"
-      ]
+      "Resource": ["${aws_s3_bucket.admin-bucket.arn}/*"]
+    },{
+      "Effect": "Allow",
+      "Action": [
```
kjdchapman: Is it bad practice to wildcard this?
`"Action": "s3:*"`

naliwajek: There are [many more permissions](https://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html) which ActiveStorage does not require.

kjdchapman: Ah, okay - sorry, heard you say "all the permissions" earlier!

PR #148, opened 2018-10-16
Run user signup and logging scheduled tasks at different times

emileswarts:
We are seeing gaps in our performance statistics reporting.
This has been an ongoing issue, and it turns out that we are running out
of memory on the cluster.

This is due to the fact that we spin up both logging and user signup
tasks at the same time.

This changes it to run logging statistics publishing at a different time to user
signup statistics publishing.


emileswarts:
These memory logs are embedded deep in CloudTrail logs.
Need to check there next time.

PR #150, opened 2018-10-23
Create Device Wifi Frontend Infrastructure for Staging

emileswarts:
This creates a new Radius frontend which has its own ECR docker
repository.

It is currently a clone of our other frontends, but I imagine that this
setup would deviate a lot as we move forward.

It has an Elastic IP associated with this frontend, which we can use to
configure our SSID to point to this to do the certificate based
authentication.

### DIFF
--- a/govwifi-frontend/outputs.tf
+++ b/govwifi-frontend/outputs.tf
@@ -1,3 +1,11 @@
 output "frontend-vpc-id" {
   value = "${aws_vpc.wifi-frontend.id}"
 }
+
+output "frontend-subnet-id" {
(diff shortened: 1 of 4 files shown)


govwifi-frontend/outputs.tf
```diff
@@ -1,3 +1,11 @@
 output "frontend-vpc-id" {
   value = "${aws_vpc.wifi-frontend.id}"
 }
+
+output "frontend-subnet-id" {
```
kjdchapman: This doesn't refer to DeviceWifi - what's it for?

PR #155, opened 2018-11-14
Spike Fargate on Logging API Staging

emileswarts:


govwifi-api/alarms.tf
```diff
@@ -126,13 +126,13 @@ resource "aws_cloudwatch_metric_alarm" "ecs-api-memory-reservation-low" {
   namespace           = "AWS/ECS"
   period              = "300"
   statistic           = "Average"
-  threshold           = "50"
+  threshold           = "40"
```
emileswarts: Since we have 1 less service running on EC2 ECS right now, this needs to decrease to keep our services running on at least 2 instances.

govwifi-api/logging-api-cluster.tf
```diff
@@ -14,19 +14,23 @@ resource "aws_ecs_task_definition" "logging-api-task" {
...
+  memory = 512
+  cpu = "256"
+  network_mode = "awsvpc"
 
   container_definitions = <<EOF
 [
     {
       "volumesFrom": [],
-      "memory": 950,
+      "memory": 512,
```
emileswarts: Our tasks don't need 950MB memory, might as well change it here as Fargate doesn't allow specific values like 950MB.

PR #158, opened 2018-11-27
Append allowed Ciphers to sshd config

kjdchapman:
My `sed-fu` is weak, but I think this does what we want.

I'd like to test this on staging first, to make sure we still have SSH clients that can connect to it. It might require some tweaking of config for clients.

### DIFF
--- a/govwifi-backend/management.tf
+++ b/govwifi-backend/management.tf
@@ -94,6 +94,7 @@ done
 IFS=$oldIFS
 
 sudo sed -i -e "s/AllowUsers ubuntu/AllowUsers $allowedUsers/g" /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


kjdchapman:
The `sed` part works, but `ssh -Q cipher` from a client still reports the unsecure CBC ciphers are available.

kjdchapman:
Ah, I'd misunderstood what `ssh -Q cipher` does - it's talking about which ciphers the _client_ can use, not the server.

Trying to connect to bastion with one of the disabled ciphers - `ssh gov-wifi-bastion-staging -c 3des-cbc`  - correctly reports that the cipher isn't available, and shares which ciphers are (correctly hard-coded to just `aes128-ctr, aes192-ctr, aes256-ctr`).

PR #162, opened 2018-12-12
Create the Infrastructure for the GovWifi Safe Restarter

emileswarts:
We need a scheduled job to execute the safe restarter every day.

Required infrastructure:

- ECR repository for the app
- Cloudwatch cron
- Task Definition
- IAM Roles and policies
- Cloudwatch Logs

### DIFF
--- a/govwifi-api/safe-restart-scheduled-task.tf
+++ b/govwifi-api/safe-restart-scheduled-task.tf
@@ -1,0 +1,195 @@
+resource "aws_cloudwatch_event_rule" "daily_safe_restart" {
+  name                = "${var.Env-Name}-daily-safe-restart"
+  description         = "Triggers at midnight Daily"
+  schedule_expression = "cron(0 0 * * ? *)"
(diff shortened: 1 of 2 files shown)


govwifi-api/safe-restart-api-cluster.tf
```diff
@@ -0,0 +1,195 @@
...
+  is_enabled          = true
+}
+
+resource "aws_cloudwatch_log_group" "safe-restart-api-log-group" {
+  name = "${var.Env-Name}-safe-restart-api-docker-log-group"
+
+  retention_in_days = 90
+}
+
+resource "aws_ecr_repository" "safe-restarter-api-ecr" {
```
kjdchapman: should this be `safe-restart-api`? Think we should stick to one of `restart` / `restarter` for this infra?

emileswarts: Yeah I can see that. Nice 1

PR #163, opened 2018-12-19
We are able to broadcast GovWifi Devices on the same SSID, so currently

emileswarts:
there is no need to have a standalone cluster to do this.

It will also save costs as we don't need to be running these unused EC2
instances.


kjdchapman:
LGTM :+1:

PR #165, opened 2019-01-02
Create scheduled task for deleting old users

emileswarts:
This will run daily at 11pm to remove inactive users to keep us GDPR
compliant

### DIFF
--- a/govwifi-api/user-signup-scheduled-tasks.tf
+++ b/govwifi-api/user-signup-scheduled-tasks.tf
@@ -96,3 +96,31 @@ resource "aws_cloudwatch_event_target" "user-signup-publish-weekly-statistics" {
 }
 EOF
 }
+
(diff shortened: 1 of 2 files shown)


govwifi-api/user-signup-scheduled-tasks.tf
```diff
@@ -96,3 +96,31 @@ resource "aws_cloudwatch_event_target" "user-signup-publish-weekly-statistics" {
...
+      subnets         = ["${var.subnet-ids}"]
+    }
+  }
+
+  input = <<EOF
+{
+  "containerOverrides": [
+    {
+      "name": "logging",
+      "command": ["bundle", "exec", "rake", "delete_users"]
```
emileswarts: I will update 'delete_users' to 'delete_inactive_users' in a separate PR

PR #170, opened 2019-01-10
Rename ecsTaskExecutionRole to be generic

emileswarts:
This needs to be re-used and we don't want it to be specific to the
logging API.


kjdchapman:
Surprised we can just change this - it won't break things that are currently using the role?

emileswarts:
Nope should be fine

PR #173, opened 2019-01-10
Unused backend variables

cybojenix:


govwifi-backend/variables.tf
```diff
@@ -34,36 +30,10 @@ variable "bastion-server-ip" {}
...
-
-variable "bastion-set-cronjobs" {}
-
-variable "ssh-key-name" {}
-
-variable "backend-instance-count" {}
-
-variable "backend-min-size" {}
-
-variable "backend-cpualarm-count" {}
```
emileswarts: Just did a quick search and it looks like this variable is still in use in
terraform/modules/govwifi-backend/modules/ecs-autoscaling/scaling-policy.tf

govwifi-backend/variables.tf
```diff
@@ -34,36 +30,10 @@ variable "bastion-server-ip" {}
...
-
-variable "backend-elb-count" {}
-
-variable "aws-account-id" {}
```
emileswarts: Same with this one:
terraform/modules/govwifi-backend/iam-roles.tf

govwifi-backend/variables.tf
cybojenix: looks like that module isn't used anymore. will remove it

govwifi-backend/variables.tf
cybojenix: realised this afterwards, it's moved to https://github.com/alphagov/govwifi-terraform/pull/173/files#diff-5766a9533fb6c3b8a8e3c3cc2bd9e8bdR9 with the other aws vars

PR #182, opened 2019-01-24
Add www as a CNAME for the admin platform

emileswarts:
This also requires setting up a new SSL certificate for the www domain.
There is another task here to redirect all requests to the www version
of the site.  This will be picked up as an isolated task as we need to
install and configure Nginx.

### DIFF
--- a/govwifi-admin/route53.tf
+++ b/govwifi-admin/route53.tf
@@ -9,6 +9,30 @@ resource "aws_route53_record" "admin" {
   }
 }
 
+resource "aws_route53_record" "admin_www" {
+  zone_id = "${data.aws_route53_zone.zone.id}"
+  name    = "www.admin.${var.Env-Subdomain}.service.gov.uk"
+  type    = "CNAME"
+  ttl     = "300"
+  records = ["admin.${var.Env-Subdomain}.service.gov.uk"]
+}
+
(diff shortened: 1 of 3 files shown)


govwifi-admin/route53.tf
```diff
@@ -9,6 +9,22 @@ resource "aws_route53_record" "admin" {
...
+  ttl     = "300"
+  records = ["admin.${var.Env-Subdomain}.service.gov.uk"]
+}
+
+resource "aws_route53_record" "docs_www" {
+  zone_id = "${data.aws_route53_zone.zone.id}"
+  name    = "www.docs.${var.Env-Subdomain}.service.gov.uk"
+  type    = "CNAME"
+  ttl     = "300"
+  records = ["docs.${var.Env-Subdomain}.service.gov.uk"]
```
cybojenix: Where is this record configured?

emileswarts: Route53

emileswarts: Or do you mean where in Terraform?
I'm not sure it looks like it was set up outside of Terraform.

emileswarts:
All our wifi.service.gov.uk dns is managed in route53 @kjdchapman

PR #189, opened 2019-02-01
Extract new user details database

emileswarts:
We are pulling the end user details table out of the main database.
This database currently has sessions and user details in it.
User details data is critical while session data is just used for
statistics.

This also introduces a new "env" variable which will replace "rack-env"
so it can be used in more contexts.


govwifi-api/user-signup-api-cluster.tf
```diff
@@ -98,6 +98,18 @@ resource "aws_ecs_task_definition" "user-signup-api-task" {
...
+          "value": "govwifi_user_${var.Env-Name}"
+        },{
+          "name": "USER_DB_PASS",
+          "value": "${var.user-db-password}"
+        },{
+          "name": "USER_DB_USER",
+          "value": "${var.user-db-user}"
+        },{
+          "name": "USER_DB_HOSTNAME",
+          "value": "${var.user-db-hostname}"
```
emileswarts: Populated but unused.
As part of this transition, the user signup api will write to both databases at the same time (for a short period of time)

kjdchapman: Wonder if there's a better name than `user_db_user` - took me a moment to work out why that's different to `db_user`.

We could rename `userdetails` and all that related stuff whilst we're here - maybe it makes more sense to talk about `accounts`, `individual_accounts`, or something like this.

Understand this confusion is temporary though, so don't feel strongly.

emileswarts: Agreed these variable names could be more descriptive.
Still feel like "user" is the correct terminology here though.

Users use GovWifi and administrators install GovWifi.
Accounts to me suggests something other than end users

govwifi-backend/db-sessions.tf
```diff
@@ -0,0 +1,66 @@
+resource "aws_db_instance" "db" {
+  count                       = "${var.db-instance-count}"
+  allocated_storage           = "${var.session-db-storage-gb}"
+  storage_type                = "gp2"
+  engine                      = "mysql"
+  engine_version              = "5.7.23"
```
kjdchapman: Don't think we should use this opportunity to update these dbs from 5.7 -> 8, but thought i'd mention the idea out loud...?

emileswarts: 👍 👍 👍

emileswarts: Oh wait this is the existing session database though, will apply to new user database

govwifi-api/user-signup-api-cluster.tf
kjdchapman: Is it a lot of churn to have `sessions_db_user` and `users_db_user`? Think that's the only name I can think of that works.

emileswarts: You mean plural instead of singular?
I've updated the variable name from user-db-user to user-db-username for clarity.

PR #198, opened 2019-02-07
Create user details read replica in Dublin

emileswarts:


emileswarts:
I didn't think this name was too bad.

It's just a variable name so can be refactored without changing anything in AWS.
What do you have in mind?

kjdchapman:
Not sure. Will think 'bout it. I liked `account` or similar, but I know you're not a fan.

My main problem (now I've thought about it more) was the side-by-side `DB_USER` and `USER_DB_USER`, which I think you've now managed to get rid of?

PR #199, opened 2019-02-07
Default to using the new user database

emileswarts:
We were writing to both, but this isn't necessary anymore as the new
database has fully taken over.


emileswarts:
ha ha

PR #201, opened 2019-02-07
Delete Read Replica for Sessions Database

emileswarts:


emileswarts:
We found that we still need the replica in London as the admin and background jobs use it

PR #204, opened 2019-02-12
Update logging scheduled task definition

emileswarts:
This needs user database details to run reporting background jobs


emileswarts:
It reads badly.  When it's grouped with USER_DB_PASSWORD and friends, it's less offensive

PR #209, opened 2019-02-18
Use ECS Fargate for the Admin

cybojenix:


govwifi-admin/auto-scaling-group.tf
```diff
@@ -1,36 +0,0 @@
-resource "aws_autoscaling_group" "ecs-admin-cluster" {
```
emileswarts: Sadly since we run migrations as ECS scheduled tasks, we will need to keep an EC2 instance around to do that.  This had to be done for the API cluster too.

cybojenix: migrates are done using `run-task`, which is supported on fargate.
See accompanying patch: https://github.com/alphagov/govwifi-admin/pull/481

PR #212, opened 2019-02-22
Add SENTRY_DSN ENV variable

tommotaylor:
Relies on https://github.com/alphagov/govwifi-build/pull/294

### DIFF
--- a/govwifi-api/variables.tf
+++ b/govwifi-api/variables.tf
@@ -56,6 +56,8 @@ variable "safe-restart-enabled" {
   default = 1
 }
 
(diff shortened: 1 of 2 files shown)


govwifi-api/variables.tf
```diff
@@ -56,6 +56,8 @@ variable "safe-restart-enabled" {
   default = 1
 }
 
+variable "safe-restart-sentry-dsn" {}
```
cybojenix: to stop the dependence on `govwifi-build`, this can safely be set to default to an empty string

PR #214, opened 2019-02-26
define govwifi-frontend security groups within the module

cybojenix:
This is a near copy + paste from the private module

Note: formatting run using `terraform fmt`


govwifi-frontend/instances.tf
```diff
@@ -37,8 +44,9 @@ DATA
 
   lifecycle {
     create_before_destroy = true
+
     ignore_changes = [
-      "user_data"
+      "user_data",
```
emileswarts: Intentional dangling comma?

cybojenix: yes, these are getting auto-applied now with `terraform fmt`

PR #217, opened 2019-02-27
create user db alarms when we have a user read-replica db

cybojenix:


cybojenix:
Discussed offline. In this case, it's fine as the variable in question is used to control whether resources relating to the read replica should be created.

PR #218, opened 2019-02-27
Environment vars required for new signup whitelist

emileswarts:
The user signup API will use our new whitelist, populate environment
variables for s3 bucket and object key


emileswarts:
Addressed in a different pr

PR #219, opened 2019-02-27
Iam whitelist

emileswarts:


govwifi-api/user-signup-api-cluster.tf
```diff
@@ -47,6 +47,12 @@ resource "aws_iam_role_policy" "user-signup-api-task-policy" {
         "s3:GetObject"
       ],
       "Resource": "arn:aws:s3:::${var.Env-Name}-emailbucket/*"
+    }, {
+      "Effect": "Allow",
+      "Action": [
+        "s3:GetObject"
+      ],
+      "Resource": ["${var.admin-bucket-arn}/*"]
```
cybojenix: we should be able to limit this to the specific file `["${var.admin-bucket-arn}/signup-whitelist.conf"]`

emileswarts: 👍

govwifi-api/user-signup-api-cluster.tf
```diff
@@ -125,6 +131,12 @@ resource "aws_ecs_task_definition" "user-signup-api-task" {
         },{
           "name": "PERFORMANCE_BEARER_COMPLETION_RATE",
           "value": "${var.performance-bearer-completion-rate}"
+        },{
+          "name": "S3_SIGNUP_WHITELIST_BUCKET",
+          "value": "govwifi-${var.rack-env}-admin"
```
cybojenix: This feels too loose to me. Perhaps we can use a data resource to tighten up:

```
"Resource": ["${data.s3_bucket.admin-bucket.arn}/signup-whitelist.conf"]
...
{"value": "${data.s3_bucket.admin-bucket.name}"}
...
data "s3_bucket" "admin-bucket" {
  name = "${var.admin-bucket-name}"
}
```

PR #220, opened 2019-02-27
NOOP Dublin trying to access the admin S3 bucket

emileswarts:
There is no user signup API in Dublin so it's fine to ignore this


cybojenix:
This has been merged as part of a different PR

PR #221, opened 2019-02-27
Rm auth email domain

emileswarts:


govwifi-api/user-signup-api-cluster.tf
```diff
@@ -60,6 +60,7 @@ EOF
 }
 
 data "aws_s3_bucket" "admin-bucket" {
+  count  = "${var.user-signup-enabled}"
```
kjdchapman: What does this do?

cybojenix: is this definitely the only thing it's being used for?

cybojenix: It prevents the data resource being created when user signup isn't enabled.
Assume this resolves the issue of querying the resource with a bucket name set to a blank string

kjdchapman: `count` / `enabled` caused the type checker in my brain to explode.

I mostly get it, just wanted to make sure this was on purpose etc.

cybojenix: Yeah, it's not always obvious.
As a flag, I prefer '-enabled' as it should amount to 1 or 0 (true/false). 
Using '-count' implies there can potentially be more than one resource.

PR #222, opened 2019-02-28
Add an alarm based on the number of user rejections we're getting

cybojenix:
While pinning it to a hard number isn't ideal, it gives us a rough alarm to work with for a while.

Large amounts of rejections can signify issues with the infrastructure, and is more of a "finger in the air" test. We'd normally expect other alarms to go off at the same time, such as a database going out of sync.

### DIFF
--- a/govwifi-api/logging-metrics.tf
+++ b/govwifi-api/logging-metrics.tf
@@ -1,0 +1,29 @@
+resource "aws_cloudwatch_log_metric_filter" "radius-access-reject" {
+  count          = "${aws_cloudwatch_log_group.logging-api-log-group.count}"
+  name           = "${var.Env-Name}-radius-access-reject"
+  pattern        = "\"\\\"authentication_result\\\": \\\"Access-Reject\\\"\""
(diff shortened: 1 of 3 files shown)


govwifi-api/logging-metrics.tf
```diff
@@ -0,0 +1,29 @@
...
+
+  metric_transformation {
+    name          = "${var.Env-Name}-radius-access-reject-count"
+    namespace     = "${local.logging_api_namespace}"
+    value         = "1"
+    default_value = "0"
+  }
+}
+
+resource "aws_cloudwatch_log_metric_filter" "radius-access-accept" {
```
kjdchapman: Is this to give us visibility of the corresponding metric - _how many `ACCESS-ACCEPT`s are we getting?_ - to help us diagnose if this alarm does go off?

cybojenix: it was originally for my ratio based one.
Was going to remove it, however it's actually a nice metric to have around

cybojenix: also yes, probably a good thing to have for diagnostic purposes + for tuning the "finger in the air" value

PR #223, opened 2019-02-28
add a global dns name for the api cluster

cybojenix:
This allows us to have a single point of entry, and let Route53 deal with what the best target is.


cybojenix:
@emileswarts, certificate generation has been added for the global domain, please re-review

PR #225, opened 2019-03-01
Use `terraform fmt` to format our configs consistently

cybojenix:


tommotaylor:
Would want to know that this does a clean `make ENV plan` on every env.

cybojenix:
All plans are clean

PR #226, opened 2019-03-05
Remove Auto Scaling Group from Ireland

emileswarts:
This ASG + EC2 instance is unused, it is required for background jobs
which happen in London.


govwifi-api/auto-scaling-group.tf
```diff
@@ -1,4 +1,5 @@
 resource "aws_autoscaling_group" "api-asg" {
+  count                     = "${var.background-jobs-enabled}"
```
cybojenix: I think you'll also need to disable the scaling policies and alarms around these too.
The launch configuration can also be dropped

emileswarts:
I can't wait for Fargate scheduled tasks to be available in London

PR #229, opened 2019-03-22
make a load balancer specific for the user signup api

cybojenix:
This is an internet facing API, so move it to a seperate load balancer.
It will allow us to lock down the other services so only our services can use them.

This is part of the move to Gov Notify.

### DIFF
--- a/govwifi-api/user-signup-api-cluster.tf
+++ b/govwifi-api/user-signup-api-cluster.tf
@@ -206,20 +206,3 @@ resource "aws_alb_target_group" "user-signup-api-tg" {
     path                = "/healthcheck"
   }
 }
-
-resource "aws_alb_listener_rule" "user-signup-api-lr" {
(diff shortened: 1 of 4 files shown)


govwifi-api/user-signup-api-cluster.tf
```diff
@@ -206,20 +206,3 @@ resource "aws_alb_target_group" "user-signup-api-tg" {
     path                = "/healthcheck"
   }
 }
-
```
Renny4Real: Just curious:
What did this do before?

cybojenix: This set up a listener rule on the main api load balancer.

When a request comes in that has the path starting with `/user-signup/`, the request will be passed over to the user-signup-api service.

This is defined by the `condition` block for setting the path, and by the `action` block for defining where it goes to

PR #238, opened 2019-04-15
Refactor security groups terraform

emileswarts:
There was a lot of duplication with the security groups being defined
next to the main files.  By moving this closer to where it is being
used, we only need to define it once.

### DIFF
--- a/govwifi-backend/outputs.tf
+++ b/govwifi-backend/outputs.tf
@@ -17,3 +17,7 @@ output "ecs-service-role" {
 output "rds-monitoring-role" {
   value = "${aws_iam_role.rds-monitoring-role.arn}"
 }
+
+output "be-admin-in" {
+  value = "${aws_security_group.be-admin-in.*.id}"
+}
--- a/govwifi-backend/outputs.tf
+++ b/govwifi-backend/outputs.tf
@@ -19,5 +19,5 @@ output "rds-monitoring-role" {
 }
 
 output "be-admin-in" {
-  value = "${aws_security_group.be-admin-in.*.id}"
+  value = "${aws_security_group.be-admin-in.id}"
 }
--- a/govwifi-backend/security-groups.tf
+++ b/govwifi-backend/security-groups.tf
@@ -1,0 +1,214 @@
+resource "aws_security_group" "be-ecs-out" {
+  name        = "be-ecs-out"
+  description = "Allow the ECS agent to talk to the ECS endpoints"
+  vpc_id      = "${aws_vpc.wifi-backend.id}"
+
+  tags {
+    Name = "${title(var.Env-Name)} Backend ECS out"
+  }
+
+  egress {
+    from_port   = 80
+    to_port     = 80
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  egress {
+    from_port   = 443
+    to_port     = 443
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  egress {
+    from_port   = 123
+    to_port     = 123
+    protocol    = "udp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  egress {
+    from_port   = 3306
+    to_port     = 3306
+    protocol    = "tcp"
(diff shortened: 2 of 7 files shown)


govwifi-backend/outputs.tf
```diff
@@ -17,3 +17,7 @@ output "ecs-service-role" {
 output "rds-monitoring-role" {
   value = "${aws_iam_role.rds-monitoring-role.arn}"
 }
+
+output "be-admin-in" {
+  value = "${aws_security_group.be-admin-in.*.id}"
```
cybojenix: there's only one resource, doesn't need a splat.

```suggestion
  value = "${aws_security_group.be-admin-in.id}"
```

govwifi-backend/security-groups.tf
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  egress {
+    from_port = 22
+    to_port   = 22
+    protocol  = "tcp"
+
+    cidr_blocks = [
+      "${split(",", var.backend-subnet-IPs)}",
+      "${split(",", var.frontend-subnet-IPs)}",
```
cybojenix: It doesn't make sense to expose the `frontend-subnet` IPs, as we aren't using VPC Peering.

`frontend-radius-IPs` covers our usage

govwifi-backend/security-groups.tf
```diff
@@ -0,0 +1,214 @@
...
+  description = "Allow inbound API calls from the RADIUS servers"
+  vpc_id      = "${aws_vpc.wifi-backend.id}"
+
+  tags {
+    Name = "${title(var.Env-Name)} Backend RADIUS API in"
+  }
+
+  ingress {
+    from_port   = 8080
+    to_port     = 8080
```
cybojenix: we exclusively use port 8443 now, though we should look into moving that to port `443` in the future

govwifi-backend/security-groups.tf
```diff
@@ -0,0 +1,214 @@
...
+
+  ingress {
+    from_port   = 8443
+    to_port     = 8443
+    protocol    = "tcp"
+    cidr_blocks = ["${split(",",var.frontend-radius-IPs)}"]
+  }
+}
+
+resource "aws_security_group" "be-sns-in" {
```
cybojenix: This is unused.
Assume it's legacy from before the `user signup` was split out

govwifi-backend/security-groups.tf
```diff
@@ -0,0 +1,214 @@
...
+
+  ingress {
+    from_port   = 80
+    to_port     = 80
+    protocol    = "tcp"
+    cidr_blocks = ["${split(",", var.amazon-sns-IPs)}", "${split(",", var.bastion-server-IP)}"]
+  }
+}
+
+resource "aws_security_group" "be-elb-out" {
```
cybojenix: This is also unused. Assume this is now part of `govwifi-frontend`

govwifi-backend/security-groups.tf
```diff
@@ -0,0 +1,214 @@
...
+
+  egress {
+    from_port   = 8080
+    to_port     = 8080
+    protocol    = "tcp"
+    cidr_blocks = ["${split(",", var.backend-subnet-IPs)}"]
+  }
+}
+
+resource "aws_security_group" "be-backend-in" {
```
cybojenix: Unused, we don't have a a load balancer for backend anymore

govwifi-backend/security-groups.tf
emileswarts: Will pick this up in a separate PR

govwifi-backend/outputs.tf
emileswarts: Nice, done

govwifi-backend/security-groups.tf
emileswarts: Remove these

govwifi-backend/security-groups.tf
emileswarts: Removed

govwifi-backend/security-groups.tf
emileswarts: Removed

govwifi-backend/security-groups.tf
emileswarts: Removed. Nice!

PR #240, opened 2019-05-02
Added bucket and policy

georgeschena:
Create bucket for product page data

### DIFF
--- a/govwifi-admin/s3.tf
+++ b/govwifi-admin/s3.tf
@@ -89,10 +89,7 @@ policy =<<POLICY
             "Principal": "*",
             "Action": [
                 "s3:GetObject",
-                "s3:GetObjectVersion",
-                "s3:PutObject",
-                "s3:PutObjectAcl",
-                "s3:PutObjectVersionAcl"
+                "s3:GetObjectVersion"
             ],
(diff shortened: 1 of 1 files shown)


govwifi-admin/s3.tf
```diff
@@ -15,6 +15,23 @@ resource "aws_s3_bucket" "admin-bucket" {
...
 }
 
+resource "aws_s3_bucket" "product-page-data-bucket" {
+  count         = 1
+  bucket        = "govwifi-${var.rack-env}-product-page-data"
+  force_destroy = true
+  acl           = "public"
+
+  tags {
+    Name        = "${title(var.Env-Name)} Product page data"
```
emileswarts: Would use var.rack-env 
var.Env-Name is "wifi" for production

govwifi-admin/s3.tf
```diff
@@ -58,3 +75,36 @@ resource "aws_s3_bucket_policy" "admin-bucket-policy" {
 }
 POLICY
 }
+
+resource "aws_s3_bucket_policy" "product-page-data-bucket" {
```
emileswarts: Probably don't want to name your bucket policy the same as the bucket with "product-page-data-bucket"

georgeschena: Good EYES

PR #245, opened 2019-05-16
Add startup scripts for the Frontend instances

cybojenix:
This moves us away from using ansible for provisioning.
The bulk of this is copied from the current API instance setup

### DIFF
--- a/govwifi-frontend/instances.tf
+++ b/govwifi-frontend/instances.tf
@@ -31,10 +31,217 @@ repo_upgrade: all
 --==BOUNDARY==
 MIME-Version: 1.0
 Content-Type: text/x-shellscript; charset="us-ascii"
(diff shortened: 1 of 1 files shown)


govwifi-frontend/instances.tf
```diff
@@ -31,10 +31,204 @@ repo_upgrade: all
...
+--==BOUNDARY==
+MIME-Version: 1.0
+Content-Type: text/x-shellscript; charset="us-ascii"
+#!/bin/bash
+# Add extra users
+
+oldIFS=$IFS
+IFS='|' read -r -a userlist <<< "${join("|", var.users)}"
+
+for credentials in "$${userlist[@]}"; do
```
cybojenix: Fairly sure we don't use this anymore (though we should be): @emileswarts

cybojenix:
Verified working on Staging

PR #247, opened 2019-05-17
Remove deprecated tasks

paroxp:
## What

We just removed the `publish_daily_statistics` rake task from the
logging api as we don't need it any more. However, the ECS scheduled
task is still running, and can't find the rake task. We need to delete
this scheduled task from the infrastructure.

This is causing throwing a RuntimeError caught by Sentry and false
alarming our Slack channel.

## How to review

- Sanity check
- Make sure the tasks are in fact removed
- Tell me off as much as possible as this is my first PR and attempt at supporting GovWifi

### DIFF
--- a/govwifi-api/logging-scheduled-tasks.tf
+++ b/govwifi-api/logging-scheduled-tasks.tf
@@ -50,31 +50,6 @@ resource "aws_iam_role_policy" "logging-scheduled-task-policy" {
 DOC
 }
 
-resource "aws_cloudwatch_event_target" "logging-publish-daily-statistics" {
-  count     = "${var.logging-enabled}"
-  target_id = "${var.Env-Name}-logging-daily-statistics"
-  arn       = "${aws_ecs_cluster.api-cluster.arn}"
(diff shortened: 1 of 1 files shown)


govwifi-api/user-signup-scheduled-tasks.tf
```diff
@@ -50,31 +50,6 @@ resource "aws_iam_role_policy" "user-signup-scheduled-task-policy" {
 DOC
 }
 
-resource "aws_cloudwatch_event_target" "user-signup-publish-daily-statistics" {
```
cybojenix: @paroxp only the logging task has been removed, this one should remain

paroxp: I thought that may be the case, but doubted due to rake target being removed?

Will resurrect. Thanks.

paroxp: Ready again.

cybojenix: it relates to a different repo.
For reference, https://github.com/alphagov/govwifi-user-signup-api/blob/master/tasks/publish_statistics.rb

PR #248, opened 2019-05-20
  Add roaming users environment variable

emileswarts:
This is the bearer token used for sending roaming statistics to the performance
  platform

### DIFF
--- a/govwifi-api/variables.tf
+++ b/govwifi-api/variables.tf
@@ -98,6 +98,10 @@ variable "performance-bearer-active-users" {
   default = ""
 }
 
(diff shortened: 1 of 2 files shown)


govwifi-api/variables.tf
```diff
@@ -98,6 +98,10 @@ variable "performance-bearer-active-users" {
   default = ""
 }
 
+variable "performance-bearer-roaming-users" {
```
cybojenix: Looks like you've not added the secrets for this variable

cybojenix: realised this isn't the wrong repo for secrets. my bad.

PR #250, opened 2019-05-23
add new task to bulk update user last logins

cybojenix:


govwifi-api/cloudwatch-events.tf
```diff
@@ -61,3 +61,12 @@ resource "aws_cloudwatch_event_rule" "daily_user_deletion_event" {
   schedule_expression = "cron(0 23 * * ? *)"
   is_enabled          = true
 }
+
+
```
emileswarts: double newline :p

govwifi-api/logging-scheduled-tasks.tf
```diff
@@ -125,6 +125,32 @@ resource "aws_cloudwatch_event_target" "logging-daily-session-deletion" {
 EOF
 }
 
+
```
emileswarts: double newline :p

PR #257, opened 2019-05-30
Update readme

steventux:
Update page title and re-add section on initialising terraform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Govwifi build
+# Govwifi terraform
 
 This repository contains instructions on how to build GovWifi end-to-end - the sites, services and infrastructure.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,7 +96,22 @@ use the `scripts/insert-secret.sh $pathandfilename` - this will add the secret t
 correct place in the password store, so it can be copied back in the required
 place.
 
-## Runnning terraform
+## Running terraform for the first time
+
+Ensure you have imported the [govwifi-build](https://github.com/alphagov/govwifi-build) sub-project:
+
+```
+make .private
```
cybojenix: This is done implicitly by anything that depends on secrets (so caught by `plan` and `init-backend`)

If we want to be explicit, can we use `make update-secrets`, which is also basically `git clone && git pull`

steventux: Ah yes, my mistake `make <env> init-backend` does it all nicely.

PR #259, opened 2019-06-04
Add user db connection details to admin app

steventux:
https://trello.com/c/ydPR0kkx/33-l-allow-gds-super-admin-to-find-user-details-of-a-username-and-vice-versa-7

Expose user db connection vars to the govwifi-admin application.

### DIFF
--- a/govwifi-admin/cluster.tf
+++ b/govwifi-admin/cluster.tf
@@ -117,6 +117,18 @@ resource "aws_ecs_task_definition" "admin-task" {
           "name": "RR_DB_NAME",
           "value": "${var.rr-db-name}"
         },{
+          "name": "USER_DB_USER",
+          "value": "${var.user-db-user}"
+        },{
+          "name": "USER_DB_PASS",
+          "value": "${var.user-db-password}"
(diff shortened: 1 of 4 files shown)


govwifi-admin/cluster.tf
```diff
@@ -116,6 +116,18 @@ resource "aws_ecs_task_definition" "admin-task" {
...
+          "value": "${var.user-db-user}"
+        },{
+          "name": "USER_DB_PASS",
+          "value": "${var.user-db-password}"
+        },{
+          "name": "USER_DB_HOST",
+          "value": "${var.user-db-host}"
+        },{
+          "name": "USER_DB_NAME",
+          "value": "${var.user-db-name}"
```
emileswarts: This could be 
```
"value": "govwifi_${var.rack-env}_users"
```

If you wanted to pass around less variables.
Not major, your call.

cybojenix: Prefer passing as a variable, as it's not the responsibility of the `admin` module to dictate the name of the user db

PR #261, opened 2019-06-10
Add script to programmatically import and trust keys for usage with pass

phantas:
'nuff said

### DIFF
--- a/scripts/trust-keys.sh
+++ b/scripts/trust-keys.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 
 cd "$(dirname "$0")"
 
-export PASSWORD_STORE_DIR=../private/passwords
+export PASSWORD_STORE_DIR=../.private/passwords
 echo "Expecting passwords under $PASSWORD_STORE_DIR"
 
 while IFS= read -r key && [[ -n "$key" ]]; do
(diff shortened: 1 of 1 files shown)


cybojenix:
File needs to be executable please

scripts/trust-keys.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd "$(dirname "$0")"
+
+export PASSWORD_STORE_DIR=../private/passwords
```
cybojenix: `.private` instead of `private`

```suggestion
export PASSWORD_STORE_DIR=../.private/passwords
```

scripts/trust-keys.sh
```diff
@@ -0,0 +1,15 @@
...
+set -euo pipefail
+
+cd "$(dirname "$0")"
+
+export PASSWORD_STORE_DIR=../.private/passwords
+echo "Expecting passwords under $PASSWORD_STORE_DIR"
+
+while IFS= read -r key && [[ -n "$key" ]]; do
+  # fetch if not already there
+  gpg --list-public-key "$key" > /dev/null || gpg --keyserver hkp://keyserver.ubuntu.com --recv-keys "$key"
```
cybojenix: throwing it out there, might also be worth falling back to the system's default keyserver as a final fallbcak:

```suggestion
  gpg --list-public-key "$key" > /dev/null ||
    gpg --keyserver hkp://keyserver.ubuntu.com --recv-keys "$key" ||
    gpg --recv-keys "$key"
```

PR #285, opened 2019-10-24
Fix frontend task definition

vixus0:
The populate-radius-certs container can't be essential since it could
fail.

There was a memory parameter missing from the container definition.

### DIFF
--- a/govwifi-frontend/iam-roles.tf
+++ b/govwifi-frontend/iam-roles.tf
@@ -78,7 +78,7 @@ resource "aws_iam_instance_profile" "ecs-instance-profile" {
 # Unused until a loadbalancer is set up
 resource "aws_iam_role_policy" "ecs-service-policy" {
   name = "${var.aws-region-name}-frontend-ecs-service-policy-${var.Env-Name}"
-  role = "${aws_iam_role.ecs-service-role.id}"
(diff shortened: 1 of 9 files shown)


govwifi-frontend/iam-roles.tf
```diff
@@ -118,13 +118,13 @@ resource "aws_iam_role" "ecs-service-role" {
...
+  name = "${var.aws-region-name}-frontend-cert-bucket-get-object-${var.Env-Name}"
+  role = "${aws_iam_role.ecs-task-role.id}"
+  policy = "${data.aws_iam_policy_document.cert_bucket_get_object.json}"
 }
 
-data "aws_iam_policy_document" "connect_to_cert_bucket" {
+data "aws_iam_policy_document" "cert_bucket_get_object" {
   statement {
     actions = [
       "s3:GetObject",
```
rjbaker: I'd be tempted to combine these read/list policies into single documents. E.g.

`
actions = [
    "s3:GetObject",
    "s3:ListObject",
]

resources = [
      "${aws_s3_bucket.frontend-cert-bucket.arn}",
      "${aws_s3_bucket.frontend-cert-bucket.arn}/*",
]
`

vixus0: Ah yeah I was wondering if that was a possibility.

PR #290, opened 2019-11-01
Add frontend NLB

sarahseewhy:
This is work in progress.

Adds main components for a Network Load Balancer in the frontend VPC:

* NLB
* Target group
* Target group attachment
* Listener

**Please especially focus on the ports and wiring.**


rjbaker:
Looks good so far. I think you'll need to create additional listeners for the UDP ports exposed by the Freeradius container here: https://github.com/alphagov/govwifi-frontend/blob/master/Dockerfile#L18

sarahseewhy:
This work is stale and Anshul and I were unable to complete the work because at the time NLB didn't support UDP.

PR #295, opened 2020-05-19
fix: add a new metrics task to log daily GovWifi users

freesteph:
The relevant task was implemented in
https://github.com/alphagov/govwifi-logging-api/pull/243 but was never
actually triggered by the environment.

### DIFF
--- a/govwifi-api/logging-scheduled-tasks.tf
+++ b/govwifi-api/logging-scheduled-tasks.tf
@@ -50,6 +50,43 @@ resource "aws_iam_role_policy" "logging-scheduled-task-policy" {
 DOC
 }
 
+resource "aws_cloudwatch_event_target" "logging-publish-daily-statistics" {
+  count     = "${var.logging-enabled}"
+  target_id = "${var.Env-Name}-logging-daily-statistics"
+  arn       = "${aws_ecs_cluster.api-cluster.arn}"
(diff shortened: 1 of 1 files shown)


govwifi-api/logging-scheduled-tasks.tf
```diff
@@ -50,6 +50,43 @@ resource "aws_iam_role_policy" "logging-scheduled-task-policy" {
...
+  arn       = "${aws_ecs_cluster.api-cluster.arn}"
+  rule      = "${aws_cloudwatch_event_rule.daily_statistics_logging_event.name}"
+  role_arn  = "${aws_iam_role.logging-scheduled-task-role.arn}"
+
+  ecs_target = {
+    task_count          = 1
+    task_definition_arn = "${aws_ecs_task_definition.logging-api-scheduled-task.arn}"
+    launch_type         = "FARGATE"
+
+    network_configuration = {
```
sarahseewhy, line 65: Out of curiosity, do we know if we need the `network_configuration` block?

freesteph: I'm afraid I don't know either; I used the other tasks as a template and substituted the bits I needed.

robinmitra: I haven't yet had the chance to investigate the network configurations myself but feel free to do this if you're able to.

robinmitra:
Feel free to get this merged in @freesteph

PR #301, opened 2020-08-06
Fix/allow logging api to write metrics

freesteph:
# Description
The logging API is the first app we're going to collect metrics from (metrics that don't go into the Performance Platform, that is) and for this purpose it needs access to the S3 metrics bucket. This just does that.


smford:
LGTM and Sarah, we see youve added landscape at the end, lets see how that goes.

PR #303, opened 2020-08-07
feat: add and schedule tasks to publish active users metrics

freesteph:
The new active users metrics are almost identical to the metrics sent
to the PerformancePlatform (i.e `publish_daily_statistics`) except
they're called `publish_daily_metrics`.

Re-use most of the previous code but use `metrics` in identifier and
schedule:

* the daily metrics generation at 2AM;
* the weekly metrics generation on Sundays at 5:45AM;
* the monthly metrics generation every 1st of the month at 6AM.

### DIFF
--- a/govwifi-api/cloudwatch-events.tf
+++ b/govwifi-api/cloudwatch-events.tf
@@ -69,3 +69,28 @@ resource "aws_cloudwatch_event_rule" "daily_gdpr_set_user_last_login" {
   schedule_expression = "cron(0 2 * * ? *)"
   is_enabled          = true
 }
+
+# new daily, weekly and monthly active users metrics published to S3
+resource "aws_cloudwatch_event_rule" "daily_active_users_metrics_event" {
+  count               = "${var.event-rule-count}"
+  name                = "${var.Env-Name}-daily-active-users-metrics-logging"
+  description         = "Triggers daily 02:00 am UTC"
+  schedule_expression = "cron(0 2 * * ? *)"
+  is_enabled          = true
+}
+
+resource "aws_cloudwatch_event_rule" "weekly_active_users_metrics_event" {
+  count               = "${var.event-rule-count}"
+  name                = "${var.Env-Name}-weekly-active-users-metrics-logging"
+  description         = "Triggers every SUN 05:45 am UTC"
+  schedule_expression = "cron(45 5 ? * 1 *)"
+  is_enabled          = true
+}
+
+resource "aws_cloudwatch_event_rule" "monthly_active_users_metrics_event" {
+  count               = "${var.event-rule-count}"
+  name                = "${var.Env-Name}-monthly-active-users-metrics-logging"
+  description         = "Triggers on the first of each month at 06:00 am UTC"
+  schedule_expression = "cron(0 6 1 * ? *)"
+  is_enabled          = true
+}
--- a/govwifi-api/logging-scheduled-tasks.tf
+++ b/govwifi-api/logging-scheduled-tasks.tf
@@ -198,6 +198,118 @@ resource "aws_cloudwatch_event_target" "gdpr-set-user-last-login" {
 EOF
 }
 
+# new metrics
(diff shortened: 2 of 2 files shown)


govwifi-api/cloudwatch-events.tf
```diff
@@ -69,3 +69,28 @@ resource "aws_cloudwatch_event_rule" "daily_gdpr_set_user_last_login" {
...
+  description         = "Triggers daily 02:00 am UTC"
+  schedule_expression = "cron(0 2 * * ? *)"
+  is_enabled          = true
+}
+
+resource "aws_cloudwatch_event_rule" "weekly_active_users_metrics_event" {
+  count               = "${var.event-rule-count}"
+  name                = "${var.Env-Name}-weekly-active-users-metrics-logging"
+  description         = "Triggers every SUN 05:45 am UTC"
+  schedule_expression = "cron(45 5 ? * 7 *)"
```
robinmitra, line 86: I believe `7` for day of week is [Saturday](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html)

robinmitra:
@freesteph Would be good to see the TF plan output as well in the PR.

govwifi-api/cloudwatch-events.tf
freesteph: All of this is copied from the other tasks :smile: 

I'll update to 1 thought, good catch.

freesteph:
> @freesteph Would be good to see the TF plan output as well in the PR.

I was going to post it but it's full of information about our cloud so I think it might be safer for people to run it locally...

govwifi-api/cloudwatch-events.tf
freesteph: actually from the cron man page:

```
       The time and date fields are:

              field          allowed values
              -----          --------------
              minute         0-59
              hour           0-23
              day of month   1-31
              month          1-12 (or names, see below)
              day of week    0-7 (0 or 7 is Sunday, or use names)
```

So will keep it for consistency

robinmitra: I'm gonna have to disagree here because I think it's important to be clear that CloudWatch uses a slightly different format to the regular cron, so we shouldn't conflate the two. And this is probably why the description of the existing task is actually incorrect (and something we should fix)!

I think @smford had fun with this in the past IIRC (or was it someone else?)!

robinmitra:
Our Terraform code is open source anyway, but if you think the output contains sensitive bits then happy to run it locally 👍🏼

freesteph:
Robin Mitra writes:

> @robinmitra commented on this pull request.
>
>
>
>> @@ -69,3 +69,28 @@ resource "aws_cloudwatch_event_rule" "daily_gdpr_set_user_last_login" {
>    schedule_expression = "cron(0 2 * * ? *)"
>    is_enabled          = true
>  }
> +
> +# new daily, weekly and monthly active users metrics published to S3
> +resource "aws_cloudwatch_event_rule" "daily_active_users_metrics_event" {
> +  count               = "${var.event-rule-count}"
> +  name                = "${var.Env-Name}-daily-active-users-metrics-logging"
> +  description         = "Triggers daily 02:00 am UTC"
> +  schedule_expression = "cron(0 2 * * ? *)"
> +  is_enabled          = true
> +}
> +
> +resource "aws_cloudwatch_event_rule" "weekly_active_users_metrics_event" {
> +  count               = "${var.event-rule-count}"
> +  name                = "${var.Env-Name}-weekly-active-users-metrics-logging"
> +  description         = "Triggers every SUN 05:45 am UTC"
> +  schedule_expression = "cron(45 5 ? * 7 *)"
>
> I'm gonna have to disagree here because I think it's important to be clear that CloudWatch uses a slightly different format to the regular cron, so we shouldn't conflate the two. And this is probably why the description of the existing task is actually incorrect (and something we should fix)!
>
> I think @smford had fun with this in the past IIRC (or was it someone else?)!

Okay cool I'll update then.

PR #304, opened 2020-08-13
Fix acm certificate config

sarahseewhy:
Add ACM configuration for `api-elb` and remove stale terraform code.

* We need to add ACM certificate and validation for api-elb because terraform fails to apply when building a new environment.
* `elb_global_cert_validation` needs to be removed because AWS now manages this automatically.


sarahseewhy:
This PR is stale. See [this PR for the update](https://github.com/alphagov/govwifi-terraform/pull/306).

PR #305, opened 2020-08-17
Add daily retrieve notifications task.

koetsier:
Run the 'retrieve_notifications' Rake task on govwifi-user-signup-api
every day at 6 AM.

Trello: https://trello.com/c/SJdoSBAm/354-spike-periodically-batch-collecting-email-and-sms-data-from-notify

### DIFF
--- a/govwifi-api/logging-scheduled-tasks.tf
+++ b/govwifi-api/logging-scheduled-tasks.tf
@@ -50,6 +50,43 @@ resource "aws_iam_role_policy" "logging-scheduled-task-policy" {
 DOC
 }
 
+resource "aws_cloudwatch_event_target" "retrieve-notifications" {
+  count     = "${var.logging-enabled}"
+  target_id = "${var.Env-Name}-retrieve-notifications"
+  arn       = "${aws_ecs_cluster.api-cluster.arn}"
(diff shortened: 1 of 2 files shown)


govwifi-api/logging-scheduled-tasks.tf
```diff
@@ -50,6 +50,43 @@ resource "aws_iam_role_policy" "logging-scheduled-task-policy" {
...
+    network_configuration = {
+      subnets = ["${var.subnet-ids}"]
+
+      security_groups = [
+        "${var.backend-sg-list}",
+        "${aws_security_group.api-in.id}",
+        "${aws_security_group.api-out.id}",
+      ]
+
+      assign_public_ip = true
```
sarahseewhy, line 74: This is more for my curiosity: why do we want set the `assign_public_ip` to true in this case?

sarahseewhy: It looks like we do it across the infrastructure so it seems fine but I was wondering if there was a reason.

sarahseewhy (approved):

govwifi-api/logging-scheduled-tasks.tf
koetsier: really no idea. I copy-pasted it from other examples

PR #306, opened 2020-08-17
Align ACM certification and Route 53 configuration with AWS

sarahseewhy:
* The project contained missing ACM certificate configuration which existed in AWS.
* We created resources for the `api-elb-regional` ACM certificate, certificate validation, and Route 53 validation record.
* We imported these resources and then applied the terraform in `staging` and `staging-london` thereby aligning Terraform code with AWS configuration.
* This solves an issue discovered in building the new staging environment where London and Ireland regions try to create the same unique Route 53 record for `api-elb-global` ACM certificate validation.
* We configured the alb_listener to compute the `api-elb-regional` ARN instead of hard coding the `elb-ssl-cert-arn` value.
* We removed the now unused `elb-ssl-cert-arn` variable from the codebase.

Paired: @smford & @sarahseewhy


smford:
This has been applied to staging and production environments.

PR #307, opened 2020-08-18
fix: correctly provision the Logging API's new scheduled tasks

freesteph:
# Description
Quoting my commit message as markdown is badly interpreting it:

```
The tasks recently introduced[1] into the logging API require two things:

1. the name of a bucket to write the new metrics into;
2. write access to that bucket.

This was effectively provisioned[2] but in the wrong place: it's not
the `logging-api-task` and `logging-api-task-policy` that need the
information but their counterpart `logging-api-scheduled-task` and
`logging-scheduled-task-policy`.

Move the variables around to make sure the tasks run swiftly.

[1]: https://github.com/alphagov/govwifi-terraform/pull/303
[2]: https://github.com/alphagov/govwifi-terraform/pull/301/
```

This should fix the `key not found: "S3_METRICS_BUCKET"` errors on Sentry.


sarahseewhy (approved):
LGTM

PR #312, opened 2020-09-11
Call the publish_(daily|weekly|monthly)_metrics rake tasks

koetsier:
These supersede the publish_*_statistics tasks and send metrics
data points to S3 instead of the the soon-to-be-retired Performance
Platform


sarahseewhy (approved):
LGTM

PR #316, opened 2020-09-17
Expose freeradius prometheus exporter endpoint

smford:
This exposes the freeradius prometheus exporter endpoint in preparation
for metrics scraping.

Pair: @sarahseewhy & @smford


sarahseewhy (approved):

sarahseewhy:
This PR was part of the work to spike introducing Prometheus and the FreeRADIUS log exporter. We implemented the spike in December, I believe this PR is now stale and can be deleted.

PR #317, opened 2020-09-18
Rename publishing events to make naming more consistent.

koetsier:
The naming of various events and handlers has become a bit ad-hoc.

This PR makes the relevant names more consistent

### DIFF
--- a/govwifi-api/cloudwatch-events.tf
+++ b/govwifi-api/cloudwatch-events.tf
@@ -70,28 +70,52 @@ resource "aws_cloudwatch_event_rule" "daily_gdpr_set_user_last_login" {
   is_enabled          = true
 }
 
-# new daily, weekly and monthly active users metrics published to S3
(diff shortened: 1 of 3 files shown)


govwifi-api/cloudwatch-events.tf
```diff
@@ -71,58 +71,58 @@ resource "aws_cloudwatch_event_rule" "daily_gdpr_set_user_last_login" {
 }
 
 # new daily, weekly and monthly active users metrics published to S3
```
alan-gds, line 73: Looks like the comment needs updating too?

alan-gds (approved):
Generally fine but please consider updating the code comment I've highlighted.

govwifi-api/cloudwatch-events.tf
koetsier: 👍 Updated it

PR #318, opened 2020-09-18
Remove Performance Platform related events

koetsier:
These events gather data for the Performance Platform and as the PP is about to be
retired we should remove these

https://trello.com/c/snhA5peo/411-remove-code-sending-data-to-the-performance-platform-only-after-data-s3-cards


robinmitra (approved):
LGTM

robinmitra:
We've decided to postpone this to a time nearer to when PP goes away.

colinbm (approved):

PR #319, opened 2020-09-22
fix: redirect status.wifi.service.gov.uk to StatusPage.io

freesteph:
Note: "govwifi2" because someone probably setup a test page some time
ago which is now inactive and has the canonical URL and we can't
access it. So "govwifi2" it is.

### DIFF
--- a/govwifi/wifi/main.tf
+++ b/govwifi/wifi/main.tf
@@ -164,7 +164,7 @@ module "dns" {
   Env-Name           = "${var.Env-Name}"
   Env-Subdomain      = "${var.Env-Subdomain}"
   route53-zone-id    = "${var.route53-zone-id}"
-  status-page-domain = "govwifi.hund.io"
+  status-page-domain = "govwifi2.statuspage.io"
 }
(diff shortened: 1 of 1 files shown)


smford:
The terraform plan indicates that some security groups are changing.  The change grants additional IPs used by AWS route53 health checks to access to the healthcheck app running on ports 8080/3000 on the radius servers.

The IPs being added are official AWS IPs from two regions: cn-northwest-1 and cn-north-1

> "52.80.197.0/25"
> "52.80.197.128/25"
> "52.80.198.0/25"
> "52.83.34.128/25"
> "52.83.35.0/25"
> "52.83.35.128/25"

The IPs where manually checked against the official AWS IP listing: https://ip-ranges.amazonaws.com/ip-ranges.json

The IP list is pulled down and consumed by terraform by this line: https://github.com/alphagov/govwifi-terraform/blob/master/govwifi-frontend/security_groups.tf#L135

smford (approved):
Looks good to me.

PR #320, opened 2020-09-23
fix: use the correct value for the StatusPage DNS entry

freesteph:
StatusPage enforces a specific value for us to set in our DNS record if we want to have it on our canonical URL. See the [configuration page](https://manage.statuspage.io/pages/bl6klm1cjshh/dns).


robinmitra (approved):
👍🏼

PR #322, opened 2020-10-26
Update policy to delete S3 objects after 30 days.

sarahseewhy:
We have a lifecycle policy to expire objects in the `wifi-emailbucket` after 30 days, but we need second policy to delete expired objects.

paired: @sarahseewhy & @camdesgov


robinmitra (approved):
🚀

PR #333, opened 2020-12-11
Autostart Prometheus Service With The Options Govwifi Needs 

camdesgov:
We want prometheus to store it's data on the EBS volume attached
to the instance itself rather than on the root volume. This is so 
the data will persist in the event that the instance is destroyed. 
We have done this by adding a new unit file to systemd.

Paired with @sarahseewhy


govwifi-prometheus/prometheus-govwifi
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Starts prometheus with the options needed by govwifi
+Documentation=https://github.com/alphagov/govwifi-terraform/blob/master/README.md
```
sarahseewhy, line 3: I'll add a not to our Prometheus post-work card to update the README about this config.

sarahseewhy (approved):
LGTM

PR #338, opened 2021-01-14
Remove IT Health Check IPs

camdesgov:
These are no longer needed now the Health Check is complete

paired with @cchani


sarahseewhy (approved):
LGTM 🚀

Thank you!

PR #339, opened 2021-01-15
Add New Ireland Prometheus Server to Staging

camdesgov:
We are adding another Prometheus server to the Dublin Region of
Govwifi AWS. This is to improve availability and to add redundancy.

Paired with @cchani

### DIFF
--- a/govwifi/staging/main.tf
+++ b/govwifi/staging/main.tf
@@ -345,7 +345,7 @@ module "govwifi-prometheus" {
   dublin-radius-ip-addresses = "${var.dublin-radius-ip-addresses}"
 
   # Feature toggle creating Prometheus server.
-  # Value defaults to 0 and should only be enabled (i.e., value = 1) in staging-london and wifi-london
+  # Value defaults to 0 and should only be enabled (i.e., value = 1)
   create_prometheus_server = 1
 
   prometheus-IPs = "${var.staging-prometheus-IPs}"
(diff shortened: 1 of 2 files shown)


govwifi/staging/main.tf
```diff
@@ -322,3 +322,31 @@ module "route53-notifications" {
...
+  fe-ecs-out    = "${module.frontend.fe-ecs-out}"
+  fe-radius-in  = "${module.frontend.fe-radius-in}"
+  fe-radius-out = "${module.frontend.fe-radius-out}"
+
+  wifi-frontend-subnet       = "${module.frontend.wifi-frontend-subnet}"
+  london-radius-ip-addresses = "${var.london-radius-ip-addresses}"
+  dublin-radius-ip-addresses = "${var.dublin-radius-ip-addresses}"
+
+  # Feature toggle creating Prometheus server.
+  # Value defaults to 0 and should only be enabled (i.e., value = 1) in staging-london and wifi-london
```
sarahseewhy, line 348: We can probably update the comments because we're adding the instance to Ireland.

sarahseewhy (approved):
LGTM, just need to update the comments so they don't go stale.

govwifi/staging/main.tf
camdesgov: Cool, I've taken out the part that refers to London but left the feature toggle information in case we need this in future

PR #340, opened 2021-01-19
Create Prometheus Server In Ireland Production

camdesgov:
Add supporting variables to ensure security groups still
function correctly.

Remove Staging Ireland Prometheus Instance. This has been
put on hold until we have a bastion instance in Ireland as
well as London. Deploying the Staging Ireland instance and
configuring it to use the London Bastion instance for ssh
connection is causing problems.


sarahseewhy (approved):
LGTM 🚀

PR #341, opened 2021-01-25
Set Storage Time For Prometheus Data To 90 Days

camdesgov:
We want to keep the Prometheus log data for a reasonable period
of time to facilitate debugging and recognise patterns. However
we don't want to keep it forever. We can easily adjust the length
of time we keep the data for in the future.

I have followed the guidance in the official documentation
on how this value should be set: https://prometheus.io/docs/prometheus/latest/storage/

Interestingly it is NOT set in the Prometheus config file. See here for more information:
https://stackoverflow.com/questions/59298811/increasing-prometheus-storage-retention

### DIFF
--- a/govwifi/staging-london/main.tf
+++ b/govwifi/staging-london/main.tf
@@ -116,7 +116,8 @@ module "backend" {
   # Whether or not to save Performance Platform backup data
   save-pp-data   = 1
   pp-domain-name = "www.performance.service.gov.uk"
-  prometheus-IPs = "${var.staging-prometheus-IPs}/32"
+  prometheus-IP-london  = "${var.prometheus-IP-london}/32"
+  prometheus-IP-ireland = "${var.prometheus-IP-ireland}/32"
 }
 
 # London Frontend ==================================================================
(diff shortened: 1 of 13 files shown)


govwifi/staging-london/main.tf
```diff
@@ -116,7 +116,8 @@ module "backend" {
   # Whether or not to save Performance Platform backup data
   save-pp-data   = 1
   pp-domain-name = "www.performance.service.gov.uk"
-  prometheus-IPs = "${var.staging-prometheus-IPs}/32"
+  prometheus-IP-london  = "${var.prometheus-IP-london}/32"
```
sarahseewhy, line 119: **Update** Never mind. I found the pull request that does this.

------------------------------------------

I think we'd need to add `prometheus-IP-london` and `prometheus-IP-ireland` to the staging-london files in`govwifi-build` for this to work when other people try to plan.

Right now those two IP values are only set for the production files.

sarahseewhy (changes requested):
There's a Terraform compilation error because the `frontend` module in `govwifi/staging/main.tf` requires the new variables:

* `prometheus-IP-london`
* `prometheus-IP-ireland`

They're being referenced in `govwifi-frontend/variables.tf` and this breaks the contract with the module declaration in `govwifi/staging/main.tf`. 

We'll need to update the [`govwifi/staging/variables.tf`](https://github.com/alphagov/govwifi-terraform/blob/d44942e84b7151df5817d4a9ae903bd498bf4ceb/govwifi/staging/variables.tf#L147) to include the new values so they can be pulled in as part of `.private`.

sarahseewhy (approved):

PR #342, opened 2021-01-27
Adds a new cron job to delete smoke test users

colinbm:
A new task has been created in the signup api to delete smoke test users. This adds a cron job to run at 23:30 after the old user deletion at 23:00

### DIFF
--- a/govwifi-api/cloudwatch-events.tf
+++ b/govwifi-api/cloudwatch-events.tf
@@ -62,6 +62,14 @@ resource "aws_cloudwatch_event_rule" "daily_user_deletion_event" {
   is_enabled          = true
 }
 
+resource "aws_cloudwatch_event_rule" "smoke_test_user_deletion_event" {
+  count               = "${var.event-rule-count}"
+  name                = "${var.Env-Name}-smoke-test-user-deletion"
+  description         = "Triggers daily 1:00 am UTC"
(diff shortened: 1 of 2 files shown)


govwifi-api/cloudwatch-events.tf
```diff
@@ -62,6 +62,14 @@ resource "aws_cloudwatch_event_rule" "daily_user_deletion_event" {
   is_enabled          = true
 }
 
+resource "aws_cloudwatch_event_rule" "smoke_test_user_deletion_event" {
+  count               = "${var.event-rule-count}"
+  name                = "${var.Env-Name}-smoke-test-user-deletion"
+  description         = "Triggers daily 1:00 am UTC"
+  schedule_expression = "cron(30 23 * * ? *)"
```
sarahseewhy, line 69: Correct me if I'm wrong, I think this cron expression says it'll run daily at 11:30PM. I know [AWS has its own cron expression configuration](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html). 

Is that intentional?

sarahseewhy (approved):

govwifi-api/cloudwatch-events.tf
colinbm: It does - I changed my mind about the time and didn't change the description - fix is there now though.

PR #345, opened 2021-02-16
Add iam users - but only on wifi-london

Ian-Nicholls:
Remove wifi from linking to accounts - so only applying the london one controls users


camdesgov (approved):
Looks good.

Tested this by running terraform plan with the branch  (no changes as expected).

PR #349, opened 2021-03-11
Importing AWS IAM items into terraform

Ian-Nicholls:
Imported the AWS IAM current state (users/groups/policy/instance profiles etc) into Terraform for management purposes.
Left to do is the attachment policy at a later date


sarahseewhy (approved):
LGTM

PR #351, opened 2021-03-16
Resolve and update cloudwatch alarms

Ian-Nicholls:
Resolved most, removed a few obsolete alarms and updated those there to match whats required.

Added a file (new_alarms.tf--) as a placeholder for alarms that need to be worked into the terraform at a later date - a lot of these are for the readreplica databases that are not currently terraformed


sarahseewhy (approved):
LGTM 🚀

PR #352, opened 2021-03-16
Wifi london grafana instance

cchani:
Added the variable for grafana server root url to the template file section of the instances.tf


Ian-Nicholls (approved):
All looking good!

PR #353, opened 2021-03-16
Add sessions trim job

colinbm:
This is a follow on from the sessions change to the admin site. As we're now storing sessions in the database we need to add a cleanup job.


Ian-Nicholls (approved):
LGTM at 08:43 am...

PR #354, opened 2021-03-17
Iam move to new nonproduct named users

Ian-Nicholls:
This stage makes sure the new, better named, users have the same permissions as the older poorly named users


sarahseewhy (approved):
LGTM

PR #356, opened 2021-03-19
Restricted google SSO domain

cchani:
Only allow the digital.cabinet-office.gov domain for SSO


Ian-Nicholls (approved):
Looking good to me

PR #360, opened 2021-03-22
Add a managed elasticsearch instance to staging backend VPC

koetsier:
The requirements for this are minimal, so we start off with a one
node instance in staging. Only inbound traffic with HTTPS is allowed
from anywhere within the subnet.ip


Ian-Nicholls (approved):
LGTM - GJ Jos :)

Might want to make the: 
     elasticsearch_version = "7.9"
a variable set somewhere instead of hardcoded(ish) in the .tf at a later date

GJ!

PR #362, opened 2021-03-24
Pin Grafana & Prometheus AMIs

camdesgov:
This is to stop the instances being continually destroyed and recreated by terraform every time a new AMI is released (this happens roughly once a week).


Ian-Nicholls (approved):
Much better idea :)

PR #365, opened 2021-03-30
Upgrade to Terraform version 0.12

koetsier:
Upgrade to 0.12. Changes include?
- new variable notation (${var.a.b.c} -> var.a.b.c)
- arrays-of-arrays are no longer considered to be arrays and so some expressions had to be modified to accommodate
- some instance counts depended on the number of other resources available. Terraform 0.12 is stricter and does not allow this, so changed it to depend on common variables.
- Some instance counts were increased to help get them in line with auto scaling


koetsier:
> image
> 
> that shouldnt be 1 there should it?
I think that is actually ok because the resource it depends on is always just one instance

Ian-Nicholls (approved):
LGTM!

PR #366, opened 2021-03-30
this should make the Grafana info persistent and not get lost in rebu…

Ian-Nicholls:
…ild using a persistent EBS

Also changed the variable passed to use _ instead of - 

And made the version of grafana docker to be set within Terraform and not hardcoded in the user_data.sh file

### DIFF
--- a/govwifi-grafana/user_data.sh
+++ b/govwifi-grafana/user_data.sh
@@ -62,9 +62,9 @@ EOF
 systemctl stop docker
 
 # format drive if needed and mount to mount point
-if [ "$(lsblk --noheadings --output FSTYPE ${grafana_drive_name})" != "${drive_format}" ]; then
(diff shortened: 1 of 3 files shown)


govwifi-grafana/user_data.sh
```diff
@@ -1,8 +1,15 @@
 #!/usr/bin/env bash
+
 set -ueo pipefail
 
 export DEBIAN_FRONTEND=noninteractive
 
+# set some vars
```
sarahseewhy, line 7: Could we make this a bit more specific?

sarahseewhy (changes requested):
All looks good, just a few comment changes to make it clear as to what the bash is doing.

sarahseewhy (approved):
LGTM

PR #380, opened 2021-04-07
made tfenv checkout the version of terraform that is in repo

Ian-Nicholls:
as it was still using version 0.11.x where as repo is now on 0.13.6 and syntax is different

prob due to setting on "macos-latest" that it uses as default would be to pull latest version


seb-szy (approved):
LGTM

PR #381, opened 2021-04-07
added description to adminips - mainly as a test to ensure tfenv set ok

Ian-Nicholls:
hopefully tf 13.6 will be used to test syntax


seb-szy (approved):
ok, let's try this approach :D

PR #388, opened 2021-04-09
made the grafana and prometheus volume tag Name not delete/add flipflop

Ian-Nicholls:
made the grafana and prometheus volume tag Name not delete/add flipflop on apply and commit the applied tf lock files


koetsier (approved):
lgtm

PR #390, opened 2021-04-09
Add Instance Level Health Checks For Grafana Server

camdesgov:
More information on the “StatusCheckFailed_System" and “StatusCheckFailed_Instance" metrics is available here: https://docs.amazonaws.cn/en_us/AmazonCloudWatch/latest/monitoring/UsingAlarmActions.html


Ian-Nicholls (approved):
LGTM - do we also need to do Service Level Health checks?

PR #394, opened 2021-04-12
Fix: Increase Cloudwatch Alert Check Interval for Grafana Target Group

camdesgov:
Metrics from the AWS/XXX namespaces only allow interval periods of 60
seconds or greater.

Also Target group and load balancer metrics only behave as expected when
 an "arn suffix" is used (they fail when you use the AWS "arn id"). Also updated.


Ian-Nicholls (approved):
LGTM

PR #397, opened 2021-04-13
Update the Bastian hosts AMI's for both eu-west-1 and eu-west-2

cchani:
Current AMI's have become outdated and running out of support. Changed the code to reflect the latest AMI's.


Ian-Nicholls (approved):
LGTM

PR #398, opened 2021-04-15
Add an Elasticsearch instance to production.

koetsier:
This commit adds an instance of Elasticsearch to the production
environment (Wifi-london).

Previously we added one to staging (#360)

This one has a similar configuration and starts off being a minimal
size as we expect minimal load.

The domain name has  had to be changed to avoid name collision between environments

Trello: https://trello.com/c/sNXzOvwe/709-provision-managed-elastic-search-in-aws-using-terraform-in-production-4


sarahseewhy (approved):
LGTM 🚀

PR #403, opened 2021-04-22
Add elasticsearch endpoint environment variable

colinbm:
## What

Add the newly created secret to Terraform

## Why

We need this in place for the rake task to use when syncing data to ES.


sarahseewhy (approved):
LGTM

PR #408, opened 2021-04-29
removed all old unused users that have not used the system in 30 days +

Ian-Nicholls:
**WHAT**

Remove all the old, unused users in AWS IAM

**WHY**

We've moved to more generic names to remove services (e.g. _jenkins_ when we are using _concourse_ so now use _pipeline_) and to tidy up unused accounts as the cleaner the IAM is the more secure/easy to manage it is


sarahseewhy (approved):
LGTM 🚀 . Thanks for owning this strand of work!

PR #409, opened 2021-04-29
also remove the policys and groups the deleted users are in

Ian-Nicholls:
**WHAT**

Remove the policy and group membership (or policy/group if only member) for the deleted users

**WHY**

To enable the users to be deleted


sarahseewhy (approved):
LGTM

PR #410, opened 2021-04-29
Remove old users and keys

Ian-Nicholls:
**WHAT**

- Remove old IAM creds from with in the govwifi-dashboard 
- Remove the servicedesk group

**WHY**

To finalise the deletion of the IAM users no longer used

**FOLLOWUP**

Do we need to look at the rest of govwifi-dashboard for removal? is it still used?


sarahseewhy (approved):
LGTM

PR #411, opened 2021-04-29
also remove defunct monitor user

Ian-Nicholls:
**WHAT**

Also remove defunct **_monitor_** user - as its not been used for 941 days (was used by other teams to monitor it loooong ago)

**WHY**

Last remaining unused old user


sarahseewhy (approved):
LGTM

PR #417, opened 2021-05-06
Make rds backup encrypted s3

Ian-Nicholls:
**WHAT**

Add new encrypted bucket, users, KMS key etc for storage of data into S3 bucket

**WHY**

Place to store RDS MySQL database backups for IT to pickup

### DIFF
--- a/govwifi-backend/s3.tf
+++ b/govwifi-backend/s3.tf
@@ -26,8 +26,8 @@ resource "aws_s3_bucket" "rds-mysql-backup-bucket" {
   acl           = "private"
 
   tags = {
-    Name   = "${title(var.Env-Name)} RDS MySQL data backup"
-    Region = title(var.aws-region-name)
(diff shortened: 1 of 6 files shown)


govwifi-backend/s3.tf
```diff
@@ -18,3 +18,48 @@ resource "aws_s3_bucket" "pp-data-bucket" {
   }
 }
 
+# Bucket to store MySQL RDS backups
+resource "aws_s3_bucket" "rds-mysql-backup-bucket" {
+  count         = var.backup_mysql_rds ? 0 : 1
+  bucket        = "${var.Env-Name}-${lower(var.aws-region-name)}-pp-data"
```
sarahseewhy: Out of curiosity what does the `pp` stand for in the bucket name?

sarahseewhy (approved):
LGTM 🚀

PR #418, opened 2021-05-06
Sync elasticsearch

cchani:
Giving the role enough permissions to access the S3 buckets for the syncing to Elasticseach


koetsier:
could you fix the commit history before merging?

sarahseewhy:
I believe this PR should be closed now. We figured out the underlying root cause and fixed in this PR: https://github.com/alphagov/govwifi-terraform/pull/429

PR #420, opened 2021-05-07
made name better (added govwifi- prefix) added alias, removed public access

Ian-Nicholls:
**WHAT**

- made name better (added govwifi- prefix)
- added KMS alias
- removed ability to have public access from S3
- enabled KMS key (could be useful!)

**WHY**

To make the [S3 bucket ticket more complete](https://trello.com/c/ZimO5nvt)


sarahseewhy (approved):
LGTM 🚀

PR #421, opened 2021-05-07
Made a new function for running targetted apply/plans with out need to modify Makefile

Ian-Nicholls:
**WHAT**

A new function in the Makefile for `terraform_targetted` that will only run the `delete-secrets` AFTER running the function

**WHY**

Previously to do this you had to manually modify the Makefile to remove the `delete-secrets` section as it needs to be there for the targetted apply/plan as the secrets need to be there for plan/apply to work. 

This would leave unencrypted secrets on the users laptop/pc and potentially leave the `delete-secrets` removed in the Makefile if forgotten to put back


koetsier (approved):
👍

PR #426, opened 2021-05-12
Remove user signup metrics scheduled tasks

koetsier:
These tasks have been moved to the logging api and are triggered
as part of those metrics tasks

See also: https://github.com/alphagov/govwifi-logging-api/pull/423

Trello: https://trello.com/c/6sgRmjNb/1084-4-update-rake-tasks-to-upload-remaining-data-points-to-elastic-search


sarahseewhy (approved):
LGTM

PR #427, opened 2021-05-12
Add govwifi-account documentation

sarahseewhy:
### What

Add README to `govwifi-account` module explaining the module purpose.

### Why

To document the fact that we did a manual import of AWS resources. We imported a bunch of IAM configuration to ensure we had a record of our AWS configuration in code. It isn't clear from the code history why we did this or that all of the resources come from a manual import.

### DIFF
--- a/govwifi-account/README.md
+++ b/govwifi-account/README.md
@@ -1,0 +1,8 @@
+# govwifi-account
+
+## Purpose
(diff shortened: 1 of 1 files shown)


Ian-Nicholls (approved):
LGTM

PR #435, opened 2021-05-17
Update Terraform target command and documentation

sarahseewhy:
### What

* Update `plan | apply` make command to accept a list of module names.
* Create new header for resource targeting and expand documentation.

### Why

Allow for easier use of Terraform targeting and ensure we've well-documented the approach.


sarahseewhy:
@Ian-Nicholls is updating the make command. I'm parking this PR while I wait for his update so the docs don't drift from the code.

Ian-Nicholls (approved):
LGTM

PR #446, opened 2021-05-26
Add more to database backup task

Ian-Nicholls:
**WHAT**

Added missing variables needed to be passed to the script (all of them!)

**WHY**

It needs them to work

### DIFF
--- a/govwifi-api/database-backup-task.tf
+++ b/govwifi-api/database-backup-task.tf
@@ -63,7 +63,7 @@ resource "aws_ecs_task_definition" "backup-rds-to-s3-task-definition" {
           "value": "${var.user-db-hostname}"
         },{
           "name": "S3_BUCKET",
-          "value": "${var.s3-bucket}"
+          "value": "govwifi-${var.Env-Name}-${lower(var.aws-region-name)}-mysql-backup-data"
         },{
           "name": "USER_DB_NAME",
           "value": "govwifi_${var.env}_users"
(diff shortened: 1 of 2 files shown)


govwifi-api/database-backup-task.tf
```diff
@@ -218,7 +263,7 @@ resource "aws_cloudwatch_event_target" "backup-rds-to-s3" {
   "containerOverrides": [
     {
       "name": "database-backup",
-      "command": ["bundle", "exec", "rake", "database_backup.sh[${var.rack-env}]"]
+      "command": ["pwd; ls -l; ./database_backup.sh"]
```
sarahseewhy, line 266: More for my curiosity than anything else, why change the command?

sarahseewhy (approved):
LGTM, just that question about the command change but I don't think it's blocking.

govwifi-api/database-backup-task.tf
Ian-Nicholls: well bundle is:
bundle. bundle - Ruby Dependency Management bundle COMMAND [--no-color] [--verbose] [ARGS]. Description. Bundler manages an application's 
and rake is:
Rake is a tool you can use with Ruby projects. It allows you to use ruby code to define "tasks" that can be run in the command line. Rake can be downloaded and included in ruby projects as a ruby gem. Once installed, you define tasks in a file named "Rakefile" that you add to your project.30 Nov 2013

(I googled!) so both are for running Ruby - this is a bash script though!

sarahseewhy: Ah, I just needed that last bit about converting the command to a bash script. I wanted to know why we moved from a rake task to bash.

PR #449, opened 2021-05-27
Update frontend ecsTaskExecutionRole names

sarahseewhy:
### What

* Remove unnecessary "SecretsManager" suffix from the govwifi-frontend `ecsTaskExecutionRole` definition. 
* Append the GovWifi cluster name to the beginning of the `ecsTaskExecutionRole` (e.g. admin, api, frontend).

### Why

This aligns the naming pattern across the three ECS clusters.

The frontend ECS cluster only has this one `ecsTaskExecutionRole`. Although this execution role is exclusively used to retrieve secrets from Secrets Manager the other ECS clusters don't follow this naming pattern. The rest are called `ecsTaskExecutionRole` without any AWS-specific service suffix.

This outlier is confusing.


Ian-Nicholls (approved):
LGTM

PR #450, opened 2021-05-27
Add missing sections tying roles+tasks

Ian-Nicholls:
**WHAT**

Added some policy tying roles and tasks similar to the tasks im cloning - an experiment to see if this works...

**WHY**

Some missing parts are in the iam section in accounts that got missed that tie the policies together - it was stopping the backup task getting access to SecretsManager

The issue I'm seeing is in AWS:
`Stopped reason ResourceInitializationError: unable to pull secrets or registry auth: execution resource retrieval failed: unable to retrieve secret from asm: service call has been retried 5 time(s): failed to fetch secret arn:aws:secretsmanager:eu-west-2:788375279931...`

### DIFF
--- a/govwifi-account/iam-roles.tf
+++ b/govwifi-account/iam-roles.tf
@@ -3558,6 +3558,37 @@ POLICY
 
 }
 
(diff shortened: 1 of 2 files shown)


sarahseewhy (approved):
Update description with a few more details about the error we're seeing in AWS. This change is a bit of an experiment to try fixing the issue.

PR #451, opened 2021-05-28
Revert "Update frontend ecsTaskExecutionRole names"

sarahseewhy:
Reverts alphagov/govwifi-terraform#449

When we applied these changes in staging it triggered our `staging-radius-cannot-connect-to-api` alert. This is unexpected behaviour that we should investigate before rolling the refactor out to production.

### DIFF
--- a/govwifi-api/iam.tf
+++ b/govwifi-api/iam.tf
@@ -1,5 +1,5 @@
 resource "aws_iam_role" "ecsTaskExecutionRole" {
-  name               = "api-ecsTaskExecutionRole-${var.rack-env}-${var.aws-region-name}"
+  name               = "ecsTaskExecutionRole-${var.rack-env}-${var.aws-region-name}"
   assume_role_policy = data.aws_iam_policy_document.assume_role_policy.json
 }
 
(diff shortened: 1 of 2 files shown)


merlian2 (approved):
LGTM

Ian-Nicholls (approved):
LGTM!

Ian-Nicholls:
I think this is because as well as making a new executionRoleARN we remove the old one that **_is currently in use_** 

old - deleted
  "executionRoleArn": "arn:aws:iam::788375279931:role/ecsTaskExecutionRole-staging-London-SecretsManager",
to new - added
  "executionRoleArn": "arn:aws:iam::788375279931:role/frontend-ecsTaskExecutionRole-staging-London",

it  would prob work fine if we ADDED the new one and left   
 "executionRoleArn": "arn:aws:iam::788375279931:role/ecsTaskExecutionRole-staging-London-SecretsManager",  
 
still there - this is prob why the old task got stuck - as we removed its executionRoleArn

so we could:

- ADD the new one - 
- apply 
- allow safe restarter to run
-  remove the old and now unused one

PR #452, opened 2021-05-28
Fix Fargate version to 1.3

koetsier:
Fargate 1.4 introduces new features that means that currently we
cannot access the Secrets manager when using version 1.4

The interim fix is to set the version to 1.3.0

See https://stackoverflow.com/questions/61265108/aws-ecs-fargate-resourceinitializationerror-unable-to-pull-secrets-or-registry


Ian-Nicholls (approved):
LGTM

PR #454, opened 2021-06-03
Get database backup script working

Ian-Nicholls:
**WHAT**

Corrected USER_ to be USERS_ for all db secrets
Using new unKMS encrypted bucket 
Corrected IAM policies to have unique SIDs and KMS privs

**WHY**

USERS_ is correct (was failing)
unKMS bucket is to get script working and then layer on the KMS
IAM policies with SIDs as that could be causing issues

### DIFF
--- a/govwifi-api/database-backup-task.tf
+++ b/govwifi-api/database-backup-task.tf
@@ -2,7 +2,7 @@ resource "aws_cloudwatch_event_rule" "backup-rds-to-s3" {
   count               = var.backup_mysql_rds ? 1 : 0
   name                = "${var.Env-Name}-backup-rds-to-s3"
   description         = "Triggers at 00:30 UTC Daily"
-  schedule_expression = "cron(30 0 * * ? *)"
+  schedule_expression = "cron(* * * * ? *)"
   is_enabled          = true
 }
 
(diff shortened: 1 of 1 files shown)


govwifi-api/database-backup-task.tf
```diff
@@ -190,6 +191,35 @@ resource "aws_iam_role_policy" "backup-rds-to-s3-task-policy" {
...
+      "Action": [
+        "s3:ListAllMyBuckets",
+        "s3:ListBucket",
+        "s3:HeadBucket"
+      ],
+      "Resource": "*"
+  }, {
+      "Sid": "sid3",
+      "Effect": "Allow",
+      "Action": [
```
sarahseewhy, line 206: You might need a ListBucket action here, I can't remember if this action is required by GetObject or PutObject. The documentation isn't clear.

sarahseewhy (approved):
LGTM. You may need that extra action but it's unclear from the documentation.

govwifi-api/database-backup-task.tf
Ian-Nicholls: thats covered here

https://github.com/alphagov/govwifi-terraform/pull/454/files#diff-fac220744d2d1cccd7923833247197c462081ff856093fe50ef30cf464a6a0dbR198-R200

sarahseewhy: Oh of course! Thank you.

PR #457, opened 2021-06-07
Increased ECS task CPU and RAM

Ian-Nicholls:
## WHAT

Beefed up the ECS tasks CPU and RAM as below (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html):

### CPU
- was - 256 (1/4 of a CPU) 
- now - 2048 (2 CPUs) 

### RAM
- was - 1024 (1 GB RAM
- now - 8192 (8 GB RAM)

## WHY

I started the backup process for prod and it was taking a long time to do nothing - these backups are a LOT larger than in Staging and need gziping and gpging inline


sarahseewhy (approved):
LGTM

PR #458, opened 2021-06-08
added the needed perms to GDS IT user so they can read S3 + KMS

Ian-Nicholls:
## WHAT

Added S3 and KMS read permissions to the backup user

## WHY

To allow them to list and backup the files


szd55gds (approved):
LGTM

PR #465, opened 2021-06-10
Refactor Grafana Secrets Manager Terraform

sarahseewhy:
### What

Move to a single credential object with multiple key/values.

- I manually created a new secret name entry `grafana/credentials` which contains all the credentials associated with Grafana.
- This has been tested by terminating and successfully recreating the staging instance.

### Why

Previously we created a single secret name entry for each of the values. This cluttered the AWS console and made it harder to find credentials. This refactor will make it easier to add credentials to services without having to create individual entries.


Ian-Nicholls (approved):
LGTM

PR #466, opened 2021-06-14
Remove versioning from RDS mysqldump backup buckets

Ian-Nicholls:
## What

Remove versioning from S3 buckets used for database backups

## Why

Not needed and makes retention policy more complicated - https://docs.aws.amazon.com/AmazonS3/latest/userguide/RemDelMarker.html


sarahseewhy (approved):
LGTM. Out of curiosity why does it make retention policy more complicated?

seb-szy (approved):
yes, we want to disable this functionality for the bucket.

PR #470, opened 2021-06-16
Remove incorrect bash command example

sarahseewhy:
### What

Remove incorrect bash command example

### Why

The example is incorrect and could confuse people. This was missed in a previous documentation update.


Ian-Nicholls (approved):
LGTM

PR #471, opened 2021-06-16
Update unencryption process and remove unused variables

sarahseewhy:
### What

Update the `unecrypt-secrets.sh` script to ignore any file named `secrets.auto.tfvars` in [govwifi-build](https://github.com/alphagov/govwifi-build/tree/master/passwords/secrets-to-copy/govwifi). These files are no longer relevant because the values are now stored in Secrets Manager. 

I have also added comments to the script because it was previously difficult to decipher. 

Remove references to the unused values from all `variables.tf` and `main.tf` files. 

Running `terraform plan` in each environment results in the following message:

```
No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
```

### Why

We have migrated to using Secrets Manager and need to update the relevant scripts to reflect this migration, as well as removing any stale code.

### DIFF
--- a/govwifi-admin/variables.tf
+++ b/govwifi-admin/variables.tf
@@ -154,9 +154,6 @@ variable "zendesk-api-endpoint" {
 variable "zendesk-api-user" {
 }
 
-variable "zendesk-api-token" {
-}
-
 variable "public-google-api-key" {
 }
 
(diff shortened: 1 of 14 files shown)


Ian-Nicholls (approved):
LGTM

camdesgov (approved):

PR #472, opened 2021-06-17
Test PR template fix

sarahseewhy:
### What

Testing to see if renaming the file and putting in `.github` enables the template (right now it's broken).

The documentation is not clear about whether the file location or name are enforced but tutorials seem to indicate it is.

### Why

The template is currently broken. The fix can only be tested once the change is incorporated into the main branch.


Ian-Nicholls (dismissed):
LGTM?

cbaines (approved):

PR #473, opened 2021-06-18
Fix production prometheus instance logging

szd55gds:
### What
* Add an IAM role, policy and instance attachment which will allow the production prometheus instance to create & write logs in Cloudwatch.
* Install awslogs to transport selected logs using the above role.
* awslogs has dependencies: python < 3.5, which requires a manual install / build from user_data.sh because there is no package in the repos for 3.4 / 3.5 or similar.
* python3.5 has many dependancies, this commit installs them in the correct sequence.

### Why
Prometheus instance has not been logging to cloudwatch since April 2021.

Link to Trello card (if applicable): https://trello.com/c/kSE54Yz9/1256-prometheus-server-is-not-logging

### Testing
Run the terraform of destiny.
Wait 12.5 mins or make tea.
Check everything re-appears.

### DIFF
--- a/govwifi-prometheus/user_data.sh
+++ b/govwifi-prometheus/user_data.sh
@@ -76,6 +76,68 @@ echo "allow 127/8" >> /tmp/chrony.conf
 mv /tmp/chrony.conf /etc/chrony/chrony.conf
 systemctl restart chrony
 
+# Inject the CloudWatch Logs configuration file contents
+sudo cat <<'EOF' > ./initial-awslogs.conf
+[general]
+state_file = /var/awslogs/state/agent-state
+
+[/var/log/syslog]
+file = /var/log/syslog
+log_group_name = ${prometheus-log-group}
+log_stream_name = {instance_id}/var/log/syslog
+datetime_format = %b %d %H:%M:%S
+
+[/var/log/auth.log]
+file = /var/log/auth.log
+log_group_name = ${prometheus-log-group}
+log_stream_name = {instance_id}/var/log/auth.log
+datetime_format = %b %d %H:%M:%S
+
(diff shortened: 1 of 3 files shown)


govwifi-prometheus/user_data.sh
```diff
@@ -76,6 +76,68 @@ echo "allow 127/8" >> /tmp/chrony.conf
...
+cd Python-3.5.9/
+sudo ./configure --enable-optimizations
+sudo make altinstall
+
+# Retrieve and run awslogs install script
+cd /
+run-until-success sudo curl https://s3.amazonaws.com/aws-cloudwatch/downloads/latest/awslogs-agent-setup.py -O
+
+# Legacy possibly - Try to circumvent pip install error with waiting for 10 seconds.
+# sleep 10
```
sarahseewhy, line 137: This is great! Quick question, is the commenting out of the `sleep` command intentional? The comment on line 136 indicates there should be a wait of 10 seconds.

sarahseewhy (approved):
All looks good to me.

Just the one clarifying question to address and fix the linting issues raised by the automatic linter and then it's good to go.

govwifi-prometheus/user_data.sh
szd55gds: yes, I've left this in place and commented 137. We are installing python differently so I don't any need for the sleep 10 now. However I left it in place in case I rediscovered a need for it, cos it's an unusual fix for something. It could be deleted if thought sensible.

sarahseewhy (approved):
LGTM 🚀

cbaines:
Changes look good, I'd suggest squashing the fixup commits in to the relevant commit.

PR #474, opened 2021-06-21
Remove unused RDS KMS key ID

sarahseewhy:
### What

Remove unused RDS KMS key ID

### Why

This variable was previously refactored out of the codebase and should have been removed from this file.


szd55gds (approved):
LGTM

PR #477, opened 2021-06-23
Update readme following onboarding

szd55gds:
### What
Update the README to assist new joiners. Clarify some options, add example environments and explain how the module name can be derived from code.

### Why
The original text was unclear and missing example environments and there was no guidance as to how to construct a module name.


Ian-Nicholls (approved):
LGTM

PR #478, opened 2021-06-24
install python3.5 and run-until-success function to fix logging

szd55gds:
### What
1. Install python3.5

2. Add the function `run-until-success` and implement for python3.5 & awslogs

### Why
Overview
* Instances have not been logging to cloudwatch since April 2021 and logging is a priority on bastion instances

Cause
* The install script for awslogs does not support python3.8
* Note: python 3.8 is the default version installed on these instances.

Python3.5
* awslogs install script only supports python < 3.5
* python3.5 is a manual install / build from user_data.sh because there is no package in the repos for 3.4 / 3.5 or similar.
* python3.5 has many dependancies, this change installs them in the correct sequence.

Function: run-until-success
* This pattern is used in prometheus and grafana modules.
* A decision has been made to replace the 'until : do : done' structure because run-until-success includes logging and can be abstracted (as a piece of tech debt).

### To Test
Run the terraform of destiny.
Wait 12.5 mins or make tea.
Check everything re-appears.

Link to Trello card (if applicable): https://trello.com/c/7ZkhOWta/1316-apply-prometheus-logging-fix-to-bastion

### DIFF
--- a/govwifi-backend/management.tf
+++ b/govwifi-backend/management.tf
@@ -177,8 +177,9 @@ EOF
 
 # Install awslogs
 # Steps required are install pre-reqs for python 3.5.n, install & build python 3.5 cos awslogs script only supports python < 3.5
-# Legacy - instance has this already - The install script requires the issue file to start with the string "Ubuntu"
-# Legacy - default - sudo echo "Ubuntu Linux 20.04 LTS - Authorized uses only. All activity may be monitored and reported. \d \t @ \n" > /etc/issue
+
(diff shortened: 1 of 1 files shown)


Ian-Nicholls (changes requested):
Probably an idea to set the version of python as a variable - as later versions of Python 3.5.x will come out (e.g. 3.5.10 is out - its EOL but they do occasionally retro fit severe security)

Ian-Nicholls (approved):
LGTM!

PR #481, opened 2021-06-28
Create KMS key for secrets CloudTrail logs

sarahseewhy:
### What

Create KMS key for encrypting Secrets Manager CloudTrail logs

Add documentation to `new-alerts.tf--` to clarify the contents are in AWS but not fully integrated into Terraform.

### Why

We're creating CloudWatch alarms which need to read from an encrypted CloudTrail log group and alert on changes in Secrets Manager. This log group needs to be encrypted and therefore needs a KMS key.

For now, we only need a single key for encrypting the logs. 

We will add the configuration for the CloudTrail and CloudWatch alarms in a subsequent PR.

[Link to Trello card](https://trello.com/c/qspTsoBf/1259-sub-task-add-security-alerting-to-secrets-manager).


Ian-Nicholls (approved):
LGTM

PR #486, opened 2021-07-05
Disable FreeRADIUS verbose logging

sarahseewhy:
### What

Disable FreeRADIUS verbose logging by changing the `radiusd-params` from `-X` (verbose logging) to `-f` (normal logging).

### Why

* It was only enabled for the DigiCert rotation which is complete.
* The FreeRADIUS server are also intermittently failing which could be related to having the verbose logging enabled.


seb-szy (approved):
All good. It Looks like a restore of the version from 7 days ago.

PR #487, opened 2021-07-06
enable grafana instances to send logs to cloudwatch

szd55gds:
### What
* Add an IAM role, policy and instance attachment which will allow the grafana instances to create & write logs in Cloudwatch.
* Install awslogs to transport selected logs using the above role.
* awslogs has dependencies: python < 3.5, which requires a manual install / build from user_data.sh because there is no package in the repos for 3.4 / 3.5 or similar.
* python3.5 has many dependancies, this commit installs them in the correct sequence.

### Why
Grafana instances have not been logging to cloudwatch since April 2021.
We need to send logs to cloudwatch, particularly in preparation for service assessment.

Link to Trello card (if applicable): https://trello.com/c/p8LL22jw/1359-fix-grafana-not-logging-to-cloudwatch

### Testing
Run the terraform of destiny.
Wait 12.5 mins or make tea.
Check everything re-appears.

### DIFF
--- a/govwifi-grafana/iam.tf
+++ b/govwifi-grafana/iam.tf
@@ -16,12 +16,12 @@ data "aws_iam_policy_document" "secrets_manager_policy" {
 }
 
 resource "aws_iam_instance_profile" "grafana-instance-profile" {
-  name       = "${var.aws-region}-${var.Env-Name}-grafana-instance-profile"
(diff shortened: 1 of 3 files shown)


sarahseewhy (approved):
LGTM 🚀

PR #490, opened 2021-07-14
Remove documentation

sarahseewhy:
### What

Remove stale documentation and documentation which has been migrated to `govwifi-dev-docs`.

### Why

We should centralise our documentation in order to have have a single source of truth. We also should remove any stale documentation since we're on-boarding an out-of-hours support team will rely more heavily on our documentation.

**Note**: we're keeping one deprecated bit of documentation for reference, but marking it as "DEPRECATED".


szd55gds (approved):
LGTM

PR #491, opened 2021-08-05
Support running terraform validate through the Makefile

cbaines:
### What
Support running terraform validate through the Makefile.

### Why
As this is useful when developing things.


sarahseewhy (approved):
Looks great, thank you! 🚀

PR #492, opened 2021-08-05
Fix the last line of numerous files

cbaines:
### What
Mostly terraform files, but a couple of markdown files as well.

### Why
Lines are not lines without the terminating `\n`. Mostly this should help with cluttering up other changes with fixing the last line of whatever file was edited.


sarahseewhy (approved):
LGTM

PR #501, opened 2021-09-02
Add Admin scheduled tasks

sarahseewhy:
### What

* Create scheduled tasks for deleting orphan users and backing up service emails.
* Create corresponding scheduled task roles, policies, and CloudWatch event rules.

### Why

We need to delete orphan users because they're cluttering up the database and making difficult to get accurate statistics on our user base.

We also need to backup service emails in the event of an AWS disaster/failure in order to reach out to our admin users quickly.

Link to Trello cards: [deleting orphan users](https://trello.com/c/M13twt1f/1510-remove-orphan-admin-users), [backing up service emails](https://trello.com/c/cofhRdNA/1517-admin-back-up-list).

### DIFF
--- a/govwifi-admin/scheduled-tasks.tf
+++ b/govwifi-admin/scheduled-tasks.tf
@@ -1,0 +1,147 @@
+resource "aws_iam_role" "scheduled_task" {
+  name = "${var.Env-Name}-admin-scheduled-task-role"
+
+  assume_role_policy = <<DOC
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "",
+      "Effect": "Allow",
+      "Principal": {
+        "Service": "events.amazonaws.com"
+      },
+      "Action": "sts:AssumeRole"
+    }
+  ]
+}
+DOC
+
+}
+
+resource "aws_iam_role_policy" "scheduled_task" {
+  name = "${var.Env-Name}-admin-scheduled-task-policy"
+  role = aws_iam_role.scheduled_task.id
+
+  policy = <<DOC
+{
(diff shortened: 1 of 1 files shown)


sarahseewhy:
We are continuing to test the change in staging. I'll mark this as a draft PR.

govwifi-admin/admin-scheduled-tasks.tf
```diff
@@ -0,0 +1,133 @@
...
+  "containerOverrides": [
+    {
+      "name": "admin",
+      "command": ["bundle", "exec", "rake", "backup:service_emails"]
+    }
+  ]
+}
+EOF
+
+}
```
cbaines, line 133: This file (and the others) are missing the trailing newline.

govwifi-admin/variables.tf
```diff
@@ -166,3 +166,13 @@ variable "bastion-ips" {
 
 variable "use_env_prefix" {
 }
+
+variable "logging-enabled" {
+  description = "Enable logging on the Admin API scheduled task. 1 means logging is enabled."
```
cbaines, line 171: The things this variable relates to don't seem logging related.

If there's the need to enable/disable the scheduled tasks, could there just be a single boolean for this (incorporating the event-rule-count variable below as well)?

colinbm: I think we assumed this was something to do with logging for the task :/

cbaines: I've just removed the variables now, they weren't used, and I wasn't aware of a need for them.

colinbm (approved):

PR #502, opened 2021-09-02
Update frontend AMIs

cbaines:
### What
Update the AMIs used in govwifi-frontend

Switch from using the Amazon Linux AMI to Amazon Linux 2, this changes
a few things, like switching to systemd as the init system for
example.

As part of this, update and try to simplify the user data. Remove the
CloudWatchMonitoringScripts as they're deprecated, and I'm not sure
what they do. Switch to using yum-cron to provide security
updates. Remove the upstart service for starting the awslogs service,
just set the configuration correctly and then start it with the
script.

### Why
The old AMIs are old.

Link to Trello card (if applicable): https://trello.com/c/lNOuXm9b/1444-story-update-amis-in-staging


szd55gds (approved):
This looks ok to me

PR #504, opened 2021-09-08
Make ECS task point to production ECR image for prod backup

sarahseewhy:
### What

Ensure database backup ECS task points to the relevant production ECR image for production backups. 

Backup task was pointing to staging ECR image instead of production. This looks like a copy/paste error which we didn't catch.

### Why

Our ECS task should point to the production ECR image. The typo isn't egregious, the production image is a promoted and relabelled version of the staging image output which we've thoroughly tested.

However, it's a typo and could be come a problematic red herring unless addressed.


cbaines (approved):

seb-szy (approved):
LGTM, I have checked ECR and there is an image with Production tag.

sarahseewhy:
Waiting to deploy to `wifi-london`.

PR #505, opened 2021-09-08
Properly use amazon-cloudwatch-agent on the frontend EC2 instances

cbaines:
### What
Previously it was just installed, but not configured/running. This
change adds configuration and gets it running. This service can
replace the awslogs service, so that's removed.

### Why
Improve the frontend instance setup, and fix sending metrics to CloudWatch.


sarahseewhy (approved):
LGTM. Thank you! 🚀

PR #506, opened 2021-09-09
Add New Secondary Staging Environment In Separate AWS Account

camdesgov:
### What

Create a new staging account in a separate AWS environment. Like the rest of our current infrastructure, the environment is split over two AWS regions (London and Ireland) and are referred to in the code as:
`staging-london-temp`
`staging-dublin-temp`
They will eventually be renamed to `staging-london` and `staging` after we have deleted the old existing staging account.

The code in `govwifi-terraform` was originally designed to run in one AWS account. To avoid breaking our production environment when this branch is merged, I’ve added a number of conditionals throughout the code (with the `is_production_aws_account` variable). 

While working on this task we also found elements in the code which we are 90% sure have been deprecated, so we have removed them from the new secondary staging account, but we have added a conditional to leave these as they are in the existing environments, just in case they really do perform some critical function of which we are unaware. 

**Changes to the Govwifi-Keys Module**
I’ve changed the public keys from being hardcoded in the `govwifi-keys` module, to being variables defined in various environments instead. This will make it easier to manage keys as we move to the model of separate accounts per environment. Directly after this code is merged to the `master` branch the `aws_key_pair.govwifi-key` will need to be re-imported into terraform to avoid terraform attempting to change it, then creating a domino effect of deleting production instances. 

The related `govwifi-build` PR should be merged directly after this one.

**Notes**

Once this code is merged to master two more tasks are required to complete the work on secondary staging:
Hooking up secondary staging to our concourse pipeline
Deleting the old staging account from our production AWS account and swapping the DNS records.

### Why

It’s best practice to keep your production and staging infrastructure in separate accounts.

### How to test

The most important thing about approving this PR is to double check that it will not break production or existing staging when it is merged. To do this run a terraform plan of this code in the following environments:
`staging-london`
`staging`
`wifi-london`
`wifi`

e.g.
`gds aws govwifi -- make staging-london plan`

You should only see the following changes: 
* ECS task definitions changes (these are adding an extra task definition variable to help sentry distinguish which environment is being used)
* Changes to they govwifi-keys module (see above for the reason).

If you spot any other changes please let me know.

As there so many commits, I recommend reviewing the diff rather than individual commits.

### Trello Card:
https://trello.com/c/Kow9cC4j/1478-story-perform-final-merge-of-main-github-branch


govwifi-admin/outputs.tf
```diff
@@ -2,3 +2,7 @@ output "db-hostname" {
   value = aws_db_instance.admin_db.address
 }
 
+output "aws-s3-admin-bucket-name" {
+  description = "the name for the admin bucket"
+  value       = aws_s3_bucket.admin-bucket[0].id
+}
```
cbaines, line 8: Missing newline here.

govwifi-backend/backup.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
```
cbaines, line 1: What's this script for?

It doesn't seem to do backups, but rather drop databases then restore them?

govwifi-frontend/cluster.tf
```diff
@@ -100,9 +108,9 @@ resource "aws_ecs_task_definition" "radius-task" {
         "name": "HEALTH_CHECK_RADIUS_KEY",
         "valueFrom": "${data.aws_secretsmanager_secret_version.healthcheck.arn}:key::"
       },{
-        "name": "HEALTH_CHECK_SSID",
-        "valueFrom": "${data.aws_secretsmanager_secret_version.healthcheck.arn}:ssid::"
-      }
+         "name": "HEALTH_CHECK_SSID",
+         "valueFrom": "${data.aws_secretsmanager_secret_version.healthcheck.arn}:ssid::"
+      }
```
cbaines, line 113: The indentation and whitespace looks better before.

govwifi/staging-dublin-temp/main.tf.old
```diff
@@ -0,0 +1,331 @@
+provider "aws" {
```
cbaines, line 1: Is this file still relevant?

govwifi/staging-london-temp/main.tf.old
```diff
@@ -0,0 +1,401 @@
+module "tfstate" {
```
cbaines, line 1: Is this file still relevant?

govwifi-backend/backup.sh
camdesgov: Good shout, I'll remove this. This is actually a work in progress to auto restore data to new staging, but it shouldn't be included here.

govwifi-admin/outputs.tf
camdesgov: Good spot, will add this

govwifi/staging-dublin-temp/main.tf.old
camdesgov: Good spot I will remove

govwifi/staging-london-temp/main.tf.old
camdesgov: Good spot I will remove

govwifi-frontend/cluster.tf
camdesgov: This is automatically indented and whitespaced by the terraform formatter

sarahseewhy (approved):
LGTM 🚀 

Let's roll this out, happy to pair.

govwifi-frontend/cluster.tf
cbaines: That may be true, but this is just one big heredoc, so the automatic formatting isn't doing anything.

I still think the previous indentation and lack of trailing whitespace is better.

sarahseewhy: Cool! Fixed and pushed in the most recent commit.

PR #507, opened 2021-09-09
Change the user for the amazon-cloudwatch-agent on frontend machines

cbaines:
### What
I just picked the cwagent user in the wizard, but it doesn't have
permission to read the relevant logs. So change the behaviour to the
default, which is the root user.

### Why
So that the amazon-cloudwatch-agent can read the log files.


szd55gds (approved):
This looks fine

PR #508, opened 2021-09-09
Revert "Make ECS task point to production ECR image for prod backup"

sarahseewhy:
### What

Reverts alphagov/govwifi-terraform#504

### Why

In the spirit of caution, we first need to create a secure copy of the back-up before we can implement this change.


sarahseewhy:
Discussing more effective options with team.

PR #512, opened 2021-09-15
Use underscores for resource names in govwifi-frontend

cbaines:
### What
Try to move towards using underscores rather than dashes, as this is
more consistent with Terraform itself. Just change the resources
within the govwifi-frontend module.

This shouldn't have any effects beyond the names used in Terraform
changing. Unfortunately, when applying this change, either the
existing resources will need to be moved (using the state mv command
for example), or need destroying and re-creating through running
terraform apply.

### Why
This is more consistent with Terraform itself.


sarahseewhy (approved):
These changes look good to me and adhere to [Terraform best practices](https://www.terraform-best-practices.com/naming).

I believe the all the changes are fine, but it could be worth double checking by running the `validate` command.

PR #513, opened 2021-09-16
Use underscores for resource names in govwifi-api module

cbaines:
### What
Try to move towards using underscores rather than dashes, as this is
more consistent with Terraform itself. Just change the resources
within the govwifi-api module.

This shouldn't have any effects beyond the names used in Terraform
changing. Unfortunately, when applying this change, either the
existing resources will need to be moved (using the state mv command
for example), or need destroying and re-creating through running
terraform apply.

### Why
Consistency with Terraform itself.


camdesgov (approved):
Look good

PR #514, opened 2021-09-22
Bring Up Grafana and Prometheus In Secondary Staging

camdesgov:
### What
Add Grafana and Prometheus to Secondary Staging London
Remove unnecessary variables for Performance platform, and migrated secret values.
Add `monitoring-stats-user` to terraform. This will need importing into the appropriate environments directly after this branch is merged with `master`. 

### Why
These changes are needed to make the Secondary Staging environment functional.

**Trello:** 
https://trello.com/c/DnFOMDcA/1586-get-grafana-prometheus-running-smoothly-in-secondary-staging

### DIFF
--- a/govwifi-frontend/route53.tf
+++ b/govwifi-frontend/route53.tf
@@ -2,10 +2,14 @@
 resource "aws_route53_record" "radius" {
   count   = var.radius-instance-count
   zone_id = var.route53-zone-id
-  name = format(
+  name = var.is_production_aws_account ? format(
     "radius%d.%s.service.gov.uk",
     var.dns-numbering-base + count.index + 1,
     var.Env-Subdomain
+    ) : format(
+    "radius%d.%s.service.gov.uk",
+    var.dns-numbering-base + count.index + 1,
+    "staging"
   )
   type    = "CNAME"
   ttl     = "300"
@@ -18,8 +22,11 @@ resource "aws_route53_record" "radius" {
 
 resource "aws_route53_health_check" "radius" {
   count = var.radius-instance-count
-  reference_name = format(
+  reference_name = var.is_production_aws_account ? format(
     "${var.Env-Name}-${var.aws-region-name}-frontend-%d",
+    count.index + 1
+    ) : format(
+    "${var.rack-env}-${var.aws-region-name}-frontend-%d",
     count.index + 1
   )
   ip_address        = element(aws_eip_association.eip_assoc.*.public_ip, count.index)
--- a/govwifi-grafana/route53.tf
+++ b/govwifi-grafana/route53.tf
@@ -11,7 +11,6 @@ resource "aws_route53_record" "grafana-route53-record" {
 }
 
 data "aws_route53_zone" "zone" {
-  name         = "wifi.service.gov.uk."
+  name         = var.is_production_aws_account ? "wifi.service.gov.uk." : "${var.Env-Subdomain}.service.gov.uk."
   private_zone = false
 }
-
--- a/govwifi-grafana/variables.tf
+++ b/govwifi-grafana/variables.tf
@@ -80,3 +80,6 @@ variable "critical-notifications-arn" {
 
 variable "use_env_prefix" {
 }
+
+variable "is_production_aws_account" {
+}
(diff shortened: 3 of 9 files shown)


Makefile
```diff
@@ -71,4 +71,4 @@ format-terraform:
 	git clone git@github.com:alphagov/govwifi-build.git .private
 
 update-secrets: .private
-	cd .private && git pull
+#	cd .private && git pull
```
cbaines: The history would be clearer without this changes, I'm guessing they were only made for testing purposes.

sarahseewhy (approved):
LGTM 🚀

Makefile
camdesgov: Hey Chris, this is an outdated change it has been removed. There are now no changes to the Makefile appearing in diff

govwifi-frontend/route53.tf
```diff
@@ -2,10 +2,14 @@
...
   zone_id = var.route53-zone-id
-  name = format(
+  name = var.is_production_aws_account ? format(
     "radius%d.%s.service.gov.uk",
     var.dns-numbering-base + count.index + 1,
     var.Env-Subdomain
+    ) : format(
+    "radius%d.%s.service.gov.uk",
+    var.dns-numbering-base + count.index + 1,
+    "STAGING"
```
cbaines, line 12: I'm not quite sure why this is capitalised? In the production case, I think Env-Subdomain would be lower case.

govwifi-grafana/route53.tf
```diff
@@ -11,7 +11,6 @@ resource "aws_route53_record" "grafana-route53-record" {
 }
 
 data "aws_route53_zone" "zone" {
-  name         = "wifi.service.gov.uk."
+  name         = var.is_production_aws_account ? "wifi.service.gov.uk." : "${var.Env-Subdomain}.service.gov.uk."
```
cbaines, line 14: I think this conditional is unnecessary, since `Env-Subdomain` is `wifi` for the production environment.

govwifi-grafana/iam.tf
```diff
@@ -67,3 +67,67 @@ resource "aws_iam_role_policy" "grafana-instance-policy" {
 EOF
 
 }
+
+resource "aws_iam_user" "monitoring-stats-user" {
```
cbaines, line 71: Could the resource name here use underscores rather than dashes?

(this is a style issue that I've been looking at recently)

govwifi-grafana/iam.tf
```diff
@@ -67,3 +67,67 @@ resource "aws_iam_role_policy" "grafana-instance-policy" {
...
+        {
+            "Sid": "AllowReadingResourcesForTags",
+            "Effect": "Allow",
+            "Action": "tag:GetResources",
+            "Resource": "*"
+        }
+    ]
+}
+EOF
+}
```
cbaines, line 133: This policy also seems to be defined within the `govwifi-account` module, so I'm not sure why it's being added here as well?

(I'm also a bit lost on how this user/policy is used, since it doesn't seem to be referenced in the Terraform)

Makefile
```diff
@@ -71,4 +71,4 @@ format-terraform:
 	git clone git@github.com:alphagov/govwifi-build.git .private
 
 update-secrets: .private
-#	cd .private && git pull
+	cd .private && git pull
```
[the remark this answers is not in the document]
cbaines: These changes have now been removed.

govwifi-frontend/route53.tf
cbaines: I think this has been changed now.

govwifi-grafana/route53.tf
cbaines: I think the relevant commit message now discusses this.

govwifi-grafana/iam.tf
cbaines: This resource has no longer been added.

govwifi-grafana/iam.tf
cbaines: This is also no longer being added.

cbaines (approved):

PR #519, opened 2021-09-27
Change module "govwifi-admin" to "govwifi_admin"

cbaines:
### What
Change module "govwifi-admin" to "govwifi_admin".

### Why
So that the resource names just have underscores rather than dashes
plus underscores.


sarahseewhy (approved):
LGTM 🚀

PR #525, opened 2021-10-04
DO NOT MERGE: refactor bastion ip related variables in govwifi backend

cbaines:
### What
*demo Pull Request for talking about Git*

Previously, the govwifi-backend module took the following variables:
bastion-server-IP, bastion-server-ip, and backend-subnet-IPs. These
changes refactor it to instead take these two variables:
bastion_server_ip and backend_subnet_cidr_blocks.

### Why
This addresses the duplication of bastion-server-ip and
bastion-server-IP, and also that these variables didn't contain IP
addresses, but instead a CIDR block.

Similarly, backend-subnet-IPs was a comma separated list of CIDR
blocks. It's been changed to backend_subnet_cidr_blocks, which is a
list of those same CIDR blocks.

The variable names and values are still confusing where the module is
used, but at least within the module things should be a bit
better. Hopefully in the future, the naming and values within
govwifi-build can be improved similarly.

# Instructions

Please have a go at reviewing this, but please use the "Start review" feature in GitHub so that it holds you comments until later. I'll tell you when to actually submit your comments.

### DIFF
--- a/govwifi-backend/management.tf
+++ b/govwifi-backend/management.tf
@@ -274,7 +274,7 @@ resource "aws_iam_instance_profile" "bastion_instance_profile" {
 resource "aws_eip_association" "eip_assoc" {
   count       = var.enable-bastion
   instance_id = aws_instance.management[0].id
-  public_ip   = replace(var.bastion_server_ip, "/32", "")
+  public_ip   = var.bastion_server_ip
 }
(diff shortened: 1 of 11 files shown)


cbaines:
Closing, as this was just used for a meeting.

PR #526, opened 2021-10-05
Cleanup govwifi backend bastion ip related variables

cbaines:
### What
Cleanup govwifi backend bastion ip related variables.

### Why
Variables are better when the name says something correct about the value.

### DIFF
--- a/govwifi-backend/management.tf
+++ b/govwifi-backend/management.tf
@@ -274,7 +274,7 @@ resource "aws_iam_instance_profile" "bastion_instance_profile" {
 resource "aws_eip_association" "eip_assoc" {
   count       = var.enable-bastion
   instance_id = aws_instance.management[0].id
-  public_ip   = replace(var.bastion-server-ip, "/32", "")
+  public_ip   = var.bastion-server-ip
 }
 
 resource "aws_cloudwatch_metric_alarm" "bastion_statusalarm" {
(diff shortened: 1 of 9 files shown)


camdesgov (approved):
LGTM. Approving on the assumption this has been, planned/tested/is non-breaking. We should probably update the variable name in `govwifi-build` to match too (https://github.com/alphagov/govwifi-build/blob/dee7a59a11e083cfc2fe3e5b6e66dddddb48326d/non-encrypted/secrets-to-copy/govwifi/staging/variables.auto.tfvars#L17). Apols if you have a PR to do this somewhere already, and I just haven't spotted it yet.

cbaines:
> LGTM. Approving on the assumption this has been, planned/tested/is non-breaking. We should probably update the variable name in `govwifi-build` to match too (https://github.com/alphagov/govwifi-build/blob/dee7a59a11e083cfc2fe3e5b6e66dddddb48326d/non-encrypted/secrets-to-copy/govwifi/staging/variables.auto.tfvars#L17). Apols if you have a PR to do this somewhere already, and I just haven't spotted it yet.

Indeed, I'm working towards making changes to the variable values/names in govwifi-build. Each time I look at them, I notice more that can just be removed though (which is also good).

Once the removing things is done, I think we can do one big update+rename change in govwifi-build, and a corresponding change in govwifi-terraform.

PR #528, opened 2021-10-05
Remove the backend-subnet-IPs-list variable

cbaines:
### What
Remove the backend-subnet-IPs-list variable

### Why
As it's unused. It's a duplication of backend-subnet-IPs. I'm also
removing this from govwifi-build.


camdesgov (approved):
Can confirm this variable looks unused.

PR #536, opened 2021-10-06
Change variables coming from govwifi-build

cbaines:
### What
This commit changes the names and many expected values of all of the
variables coming from govwifi-build, for all 6 current
environments. It's intended to be merged alongside a corresponding
change in govwifi-build which changes the names and many variable
values.

### Why
There are two things I'm trying to improve with this commit. Variable
names (underscores rather than hyphens), and having the variable name
accurately describe the values (so bastion_server_ip should be an IP
address for example).

To avoid needing to coordinate many merges between govwifi-build and
govwifi-terraform, I've grouped this all together in one large
commit. Hopefully such a change won't be necessary again.


cbaines:
https://github.com/alphagov/govwifi-terraform/pull/537 should be merged before this, and then this branch should be rebased on master.

cbaines:
There are lots of meaningless changes, but I've noted some of the things below that it's really nice to get rid of :tada: 

No more adding /32 to IP addresses, then calling the result an IP address:
```terraform
grafana-IP            = "${var.grafana-IP}/32"
```

No more lists of CIDR blocks hiding as lists of IP addresses:
```terraform
cidr_blocks = var.administrator-IPs
```

No more trying to explain why variable names lie about the value.
```terraform
  # The value of bastion-server-IP isn't actually an IP address, but a
  # /32 CIDR block, extract the IP address from CIDR block here before
  # passing it on.
  bastion_ip = split("/", var.bastion-server-IP)[0]
```

No more mysterious splitting on something that doesn't contain commas.
```terraform
  prometheus-IPs = concat(
    split(",", "${var.prometheus-IP-london}/32"),
    split(",", "${var.prometheus-IP-ireland}/32")
  )
```

PR #543, opened 2021-10-13
Try out tflint

cbaines:
### What
Add some initial configuration for tflint.

### Why
A linter might be a way of trying to improve and maintain the quality of the Terraform configuration.


cbaines:
I've raised https://github.com/alphagov/govwifi-terraform/pull/544 which I think removes the unused variables tflint is finding.

PR #552, opened 2021-10-18
Neaten up AWS providers

cbaines:
### What
Specify required providers for all modules, and rename the `aws.route53-alarms` and `aws.AWS-main` providers.

### Why
This addresses some warnings I've seen from Terraform, and some stylistic thoughts I had.

Link to Trello card: https://trello.com/c/5OosQMdJ/1676-neaten-up-the-providers-in-terraform

### DIFF
--- a/govwifi/staging-dublin-temp/main.tf
+++ b/govwifi/staging-dublin-temp/main.tf
@@ -1,6 +1,6 @@
 module "tfstate" {
   providers = {
-    aws = aws.AWS-main
+    aws = aws.main
   }
 
   source             = "../../terraform-state"
@@ -38,7 +38,7 @@ terraform {
 }
 
 provider "aws" {
-  alias  = "AWS-main"
+  alias  = "main"
   region = var.aws-region
 }
 
@@ -50,7 +50,7 @@ provider "aws" {
 # Backend ==================================================================
 module "backend" {
   providers = {
-    aws = aws.AWS-main
+    aws = aws.main
   }
 
   source                    = "../../govwifi-backend"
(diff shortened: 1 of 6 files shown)


sarahseewhy:
> I put down the motivation (why) in the individual commits, and summarised it in the Pull Request description. I can copy this in to the Pull Request description if you think it's useful, but that does seem more confusing to me than just keeping this information associated with the relevant commits?

This could be a difference of perspective on raising PRs and I'm happy to discuss in more detail.

For someone who hasn't been involved in the work, ideally the trello card is the entry point for understanding what the change is, why we're doing it, and the value it brings.

If the card is sparse on details, the PR description is the next step to understanding the change with the commits telling the longer story. For me, it's not immediately clear what the motivation for the change is in the PR description or the commits, even though they do provide a bit more detail as to what has changed. 

The description refers to some stylistic changes but not why the changes are required or their benefits. That information may seem intuitively obvious, especially when you've been doing the work, but it may not be as obvious to someone who is coming to this for the first time or in a year when we're trying to figure out why this change occurred.

Why is it a good idea to add `required_providers` to various child modules? There are [different ways to declare providers in Terraform](https://stackoverflow.com/questions/68216074/do-terraform-modules-need-required-providers), we've historically chosen one way (inheritance) and now we're moving in a different direction, it'd be helpful to over-communicate the thinking behind this decision

cbaines:
> Thanks for adding the documentation to the commit. I appreciate this is a bit frustrating, commit and PR description clarity is subjective and I tend err on pushing back on anything that isn't immediately clear to me when I come to something for the first time.
> 
> I'm passionate about ensuring we over-communicate motivation in commits in PRs so we can improve the experience if interacting with the repo.

No problem, I think code review is the best time to discuss changes in detail, and that the most important parts of commit messages is the motivation/why for the changes so it's good to have these discussions.

PR #556, opened 2021-10-19
Manage the admin app data s3 bucket using a terraform output

cbaines:
### What
Rather than hardcoding the name of the bucket in the Terraform for
each environment, pass the value around instead. Additionally, change some of the variable naming involved.

### Why
Less hardcoded values plus more informative variable names should make the Terraform easier to read and understand.

Link to Trello card: https://trello.com/c/eTobuYyV/1684-convert-hard-coded-bucket-name-to-output


cbaines:
Just opening this as a draft at the moment, as I haven't fully validated if this approach will work. It's also the first use of the Terraform remote state data source, so that would be good to discuss.

PR #584, opened 2021-11-01
Make data export bucket publicly accessible [wait for Cyber 👍 before merging]

koetsier:
### What
make govwifi-<environment>-export-data-bucket buckets publicly readable

### Why
Our performance metric data needs to be made public through data.gov.uk
which means that the S3 bucket containing our data must be made
publicly readable.

Link to Trello card (if applicable): https://trello.com/c/7XXcjgKz/1645-write-aws-lambda-to-create-metrics-file-in-s3


sarahseewhy (approved):
LGTM

PR #588, opened 2021-11-04
Allow logging_api_task_policy ecs task to write to export-data-bucket

koetsier:
### What
Allow logging_api_task_policy ecs task to write to export-data-bucket S3 bucket

### Why
The scheduled sync_s3_to_data_bucket task requires write access to the export-data-bucket S3 bucket


sarahseewhy (approved):
🚀

PR #592, opened 2021-11-08
Perform most remaining variable and output name changes

cbaines:
### What
Perform most remaining variable and output name changes

### Why
The direction is to converge on using underscores rather than dashes
in variable names, this is more consistent with Terraform itself.

Link to Trello card: https://trello.com/c/aY5JXh8R/1783-perform-most-remaining-variable-and-output-name-changes


sarahseewhy (approved):
🚀

PR #593, opened 2021-11-08
Rename some locals

cbaines:
### What
Use underscores for locals names, and amend the README.md to mention this when talking about underscores.

### Why
This is more consistent with Terraform itself, and something that the tflint naming plugin checks.

Link to Trello card: https://trello.com/c/XuYh7Hg9/1784-use-underscores-for-locals-names-in-govwifi-terraform


sarahseewhy (approved):
LGTM.

PR #596, opened 2021-11-08
[DO NOT MERGE]Remove references to "temp" in staging

camdesgov:
### What
Update staging main.tf and variables.tf files to remove references to temp
Leave the key and directory names as they are for
now, they will be changed in a later commit to avoid breaking things.

### Why
The "temp" suffix was a useful distinguisher when we had two staging environments, it is now no longer necessary.

Link to Trello card (if applicable): https://trello.com/c/LZDJ09Om/1776-update-references-to-staging-temp


govwifi/staging-dublin-temp/main.tf
```diff
@@ -50,7 +50,7 @@ data "terraform_remote_state" "london" {
   backend = "s3"
 
   config = {
-    bucket = "govwifi-staging-temp-london-tfstate"
+    bucket = "govwifi-staging-london-tfstate"
```
cbaines, line 53: Is this something you're intending to change? I'm a bit confused as to whether this bit of the commit message applies:
> Leave the key, S3 terraform state bucket and directory names as they are for now

Also, I'm not sure what the "key" is?

camdesgov: > Is this something you're intending to change? I'm a bit confused as to whether this bit of the commit message applies:
> 
> > Leave the key, S3 terraform state bucket and directory names as they are for now
> 
> Also, I'm not sure what the "key" is?

Ah this was something I changed without meaning to (I've now changed it back and force pushed). Key refers to the KMS/ssh key names. I've edited the commit message to make this a bit clearer.

cbaines: :+1:

cbaines:
I think the changes on this branch look reasonable, are you able to rebase to get rid of the merge commit?

PR #597, opened 2021-11-08
Staging cleanup

camdesgov:
### What
Remove conditional from govwifi-keys module, this is no longer needed.
Update email bucket name to use subdomain in new environments.
Update bucket name to be dynamic rather than hardcoded

### Why
Describe why the change was necessary
AWS requires the bucket name to be unique. Using the subdomain instead of the
environment name ensures this always happens.

https://trello.com/c/LZDJ09Om/1776-update-references-to-staging-temp

### DIFF
--- a/govwifi-emails/s3sns.tf
+++ b/govwifi-emails/s3sns.tf
@@ -1,6 +1,6 @@
 # S3 bucket to store the emails
 resource "aws_s3_bucket" "emailbucket" {
-  bucket        = "${var.env_name}-emailbucket"
+  bucket        = var.is_production_aws_account ? "${var.env_name}-emailbucket" : "${var.env_subdomain}-emailbucket"
   force_destroy = true
 
   policy = <<EOF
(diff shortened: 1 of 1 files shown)


cbaines:
For the govwifi-emails related commit, I'd expect Terraform to be raising an error if it couldn't create a bucket since one already exists with that name? I don't see anything relevant in the plan for either environment which uses the govwifi-emails module.

As for the second change, I've already got a Pull Request open for that https://github.com/alphagov/govwifi-terraform/pull/590

sarahseewhy (approved):
LGTM. Approving now so you can merge when you're ready.

cbaines:
> For the govwifi-emails related commit, I'd expect Terraform to be raising an error if it couldn't create a bucket since one already exists with that name? I don't see anything relevant in the plan for either environment which uses the govwifi-emails module.

What happened to discussing this?

PR #607, opened 2021-11-10
Fix a few issues with the composite alarms

cbaines:
### What
Fix an output and the staging environments config.

### Why
To get Terraform to run correctly.

Link to Trello card: https://trello.com/c/RmuNbaKm/1789-manage-the-composite-alarms-with-terraform


szd55gds (approved):
LGTM

PR #608, opened 2021-11-11
Redirect HTTP to HTTPS for the admin site

cbaines:
### What
Redirect HTTP to HTTPS for the admin site

I noticed this wasn't a feature when looking at why the smoke tests
were failing on staging. The smoke tests first try to access the admin
site over http, and this was failing as you'd expect given the load
balancer doesn't listen on port 80.

### Why
I've got no idea why the smoke tests were passing in production, but I
think it's worthwhile to support redirecting from HTTP to HTTPS both
for the smoke tests and for any users accessing the admin site first
over HTTP.

Link to Trello card: https://trello.com/c/t4cZcYfr/1807-investigate-why-staging-smoke-tests-are-failing

### DIFF
--- a/govwifi-admin/loadbalancer.tf
+++ b/govwifi-admin/loadbalancer.tf
@@ -28,3 +28,18 @@ resource "aws_alb_listener" "alb_listener" {
   }
 }
 
+resource "aws_alb_listener" "http" {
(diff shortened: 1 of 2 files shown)


szd55gds (approved):
LGTM

PR #612, opened 2021-11-19
Add Pipeline related IAM Resources To New Staging

camdesgov:
### What
These IAM resources had been created manually in the new staging environment.
This commit adds them to terraform. This is also the first stage of migrating
all IAM resources out of the govwifi-account module and including them in their
related modules (e.g. govwifi-backend). These resources have been added with a
conditional so they are only created by the govwifi-api module in staging (this
avoids duplication in production).

### Why
All resources should be in terraform.


koetsier (approved):
LGTM 🥇

PR #614, opened 2021-11-22
Cleanup: Remove `use_env_prefix` conditional

camdesgov:
### What
Cleanup: Remove `use_env_prefix` conditional

### Why
The code is no longer necessary since the naming convention between the primary (production) and secondary (staging) AWS accounts are the same. The `use_env_prefix` variable was originally added to distinguish between secrets when our staging and production environments were in the same AWS account (they are now in separate AWS Accounts).

Testing: 
This should produce no changes when terraform plan is run in staging or production.

Trello:
https://trello.com/c/SfNTpuMo/1819-cleanup-remove-useenvprefix-from-terraform

### DIFF
--- a/govwifi-admin/secrets-manager.tf
+++ b/govwifi-admin/secrets-manager.tf
@@ -3,7 +3,7 @@ data "aws_secretsmanager_secret_version" "notify_api_key" {
 }
 
 data "aws_secretsmanager_secret" "notify_api_key" {
-  name = var.use_env_prefix ? "staging/admin-api/notify-api-key" : "admin-api/notify-api-key"
+  name = "admin-api/notify-api-key"
 }
 
 data "aws_secretsmanager_secret_version" "zendesk_api_token" {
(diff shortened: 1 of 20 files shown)


cbaines:
I'm guessing this commit only does part of the required changes, right? The secrets in staging presumably need to be renamed.

camdesgov:
> I'm guessing this commit only does part of the required changes, right? The secrets in staging presumably need to be renamed.

No renaming necessary, just deletion. After this PR is merged I will then have to remove everything from our AWS production account Secrets Manager that begins with `staging/`, to complete the clean up. The secrets in the new staging account never had the `staging/` prefix. Let me know if that doesn't makes sense, and I can chat you through it on a GMeet.

cbaines:
> > I'm guessing this commit only does part of the required changes, right? The secrets in staging presumably need to be renamed.
> 
> No renaming necessary, just deletion. After this PR is merged I will then have to remove everything from our AWS production account Secrets Manager that begins with `staging/`, to complete the clean up. The secrets in the new staging account never had the `staging/` prefix. Let me know if that doesn't makes sense, and I can chat you through it on a GMeet.

Ah, I think I see where my thinking is wrong, I was assuming that the new staging environments used the `use_env_prefix` setting, but they don't, the old staging environments did. So this is just removing unused Terraform.

cbaines (approved):

PR #621, opened 2021-11-24
Convert some variables to locals

cbaines:
### What
Convert some variables to locals

### Why
For three variables that don't (and shouldn't) vary. This commit
covers wifi-london and both staging environments, wifi was already
changed in ca71b7b0c9dc4f673d0c27e403b6c05af3ffa24f.

Variables form part of the interface for Terraform configuration, and
as is in the name, they're usually used for things that vary. The
env_name, env_subdomain and product_name for the wifi environment
Terraform configuration aren't things that vary, so I think it's
clearer to make them locals, rather than variables.

Link to Trello card: https://trello.com/c/x0yzOMIW/1854-convert-more-non-varying-variables-to-locals-in-govwifi-terraform

### DIFF
--- a/govwifi/staging-dublin/locals.tf
+++ b/govwifi/staging-dublin/locals.tf
@@ -1,4 +1,11 @@
 locals {
+  env_name      = "staging"
+  env_subdomain = "staging.wifi" # Environment specific subdomain to use under the service domain
+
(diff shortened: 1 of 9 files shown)


szd55gds (approved):
This is sensible and LGTM

PR #630, opened 2021-11-29
Change how subnets are handled in govwifi-frontend

cbaines:
### What
Rather than passing in the zone_count, zone_names and zone_subnets,
lookup the availability zone details from AWS, and derive the subnet
blocks for the VPC CIDR block.

These are basically the same changes as
b3e62eebe8edcb35cb039fda37fec8cb554f2b40, plus removing the
environment level zone_ variables, since they're no longer needed.

Applying this without re-creating the subnets will require moving the
old subnets to the new places in the statefiles.

### Why
I think there are a few advantages to this. There are less variables
to manage, which simplifies the configuration, instead values looked
up or derived from the appropriate place. This is important in
Terraform because it informs resource dependencies.

Link to Trello card: https://trello.com/c/dom8wPSj/1868-change-how-subnets-are-handled-in-govwifi-frontend-in-govwifi-terraform

### DIFF
--- a/govwifi-frontend/frontend.tf
+++ b/govwifi-frontend/frontend.tf
@@ -26,17 +26,17 @@ resource "aws_route" "internet_access" {
   gateway_id             = aws_internet_gateway.wifi_frontend.id
 }
 
-# CREATE SUBNET IN EACH AZ
+data "aws_availability_zones" "zones" {}
 
 resource "aws_subnet" "wifi_frontend_subnet" {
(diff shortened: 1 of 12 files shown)


szd55gds (approved):
I pulled this code and ran a 'plan' against Dublin Staging. It appeared to work ok suggesting a number of related subnet changes and zero errors. So I will approve this PR on that basis. Sounds sensible, LGTM.

PR #635, opened 2021-11-30
Drop the min_capacity for the Logging API down to 2

cbaines:
### What
Drop the min_capacity for the Logging API down to 2

### Why
Now that the Logging API is running with more threads, I think it can
process more requests in parallel, removing the need to run so many
tasks.

Link to Trello card: https://trello.com/c/0Qjq8a5b/1879-drop-the-mincapacity-for-the-logging-api-down-to-2


szd55gds (approved):
I've been closely involved with this and it sounds sensible. LGTM.

PR #636, opened 2021-11-30
Add extra documentation around access logs bucket setup

camdesgov:
### What
Add extra documentation around access logs bucket setup.

### Why
Lack of documentation caused confusion during a recent disaster recovery exercise. In the future
we would like to look at removing the need to specify particular regions for resources entirely.

https://trello.com/c/BdeRMbFs/1820-practice-bringing-up-a-new-account-from-scratch-in-dr-account

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,12 +123,10 @@ It should look like this, `module.backend.aws_instance.management`:
 | module  | backend | aws_instance | management |
 
 ## Bootstrapping terraform
+You will need to run terraform in the **eu-west-2 (London) region** first. If you try to run terraform in the eu-west-1 region first, you will encounter errors because the Dublin Terraform looks up outputs from the London region statefile.
 
 ### Setting Up Remote State
-We use remote state, but there is a chicken and egg problem of creating
-a state bucket in which to store the remote state. When you are first creating a
-new environment (or migrating an environment not using remote state to use
-remote state) you will need to do the following
+We use remote state, but there is a chicken and egg problem of creating a state bucket in which to store the remote state. When you are first creating a new environment (or migrating an environment not using remote state to use remote state) you will need to do the following
 
 Comment out the section
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,7 +136,12 @@ terraform {
   backend          "s3"             {}
 }
 ```
-in the main.tf file of the environment to be migrated
+in the main.tf file of the environment to be migrated. Then comment out the lines related to replication configuration in govwifi-terraform/terraform-state/accesslogs.tf :
+```
+replication_configuration{
+  ....
+}
+```
```
cbaines, line 144: I think we should say why it needs to be commented out, as it's probably not obvious.

cbaines: Also, I believe the `replication_configuration` in `tfstate.tf` causes the same problem as that in `accesslogs.tf`.

README.md
```diff
@@ -123,22 +123,26 @@ It should look like this, `module.backend.aws_instance.management`:
 | module  | backend | aws_instance | management |
 
 ## Bootstrapping terraform
+You will need to run terraform in the **eu-west-2 (London) region** first. If you try to run terraform in the eu-west-1 region first, you will encounter errors.
```
cbaines, line 126: Personally I'd put "because the Dublin Terraform looks up outputs from the London region statefile", but encounter errors is OK as well.

README.md
```diff
@@ -164,7 +168,7 @@ Then run
 make <ENV> apply
 ```
 
-This should then copy the state file to s3 and use this for all operations
+This should then copy the state file to s3, which will be used for all operations. Once you have run terraform in both regions, and both the S3 buckets used for the access log replication have been created, uncomment the section in govwifi-terraform/terraform-state/accesslogs.tf.
```
cbaines, line 171: Would be good to mention the state bucket here as well

cbaines (approved):
I've made a few more comments, but I'm also fine approving :+1:

PR #639, opened 2021-12-01
Fix an issue with the Logging API task definition

cbaines:
### What
Fix an issue with the Logging API task definition

### Why
The change in 2e68072df39f596175aff35fc0e54b8c0453f25c was wrong, the
VOLUMETRICS_ENDPOINT value can't be valueFrom, since it's now not a
secret.


camdesgov (approved):
lgtm

PR #640, opened 2021-12-01
Drop the min_capacity for the Authentication API down to 2

cbaines:
### What
Drop the min_capacity for the Authentication API down to 2

### Why
Now that the Authentication API is running with more threads, I think
it can process more requests in parallel, removing the need to run so
many tasks.

Link to Trello card: https://trello.com/c/80hoHxer/1881-use-more-threads-for-the-authentication-api-and-less-tasks


szd55gds (approved):
LGTM

PR #641, opened 2021-12-02
Remove most of is production aws account

cbaines:
### What
Remove most of `is_production_aws_account`.

There's one use in the `govwifi-api` module, I'll tackle that separately as it involves introducing some environment specific name for the wordlist bucket. There's another use in `govwifi-frontend` for naming the frontend certs buckets.

### Why
This was introduced as a temporary thing as part of the migration to staging in a separate AWS account. Now that this migration is near to finishing, it can be removed (and this will make the Terraform clearer).


govwifi-frontend/certs.tf
```diff
@@ -1,6 +1,6 @@
 resource "aws_s3_bucket" "frontend_cert_bucket" {
   count  = 1
-  bucket = var.is_production_aws_account ? "govwifi-${var.rack_env}-${lower(var.aws_region_name)}-frontend-cert" : "govwifi-${var.env_subdomain}-${lower(var.aws_region_name)}-frontend-cert"
+  bucket = "govwifi-${var.rack_env}-${lower(var.aws_region_name)}-frontend-cert"
```
camdesgov, line 3: I think it would be better to use the `env_subdomain` value instead of the `rack_env` and update the bucket names in manually production. The subdomain name will always be unique, where as the rack_env is only ever "staging" or "production" and could cause "bucket already exists" errors if we ever try to terraform a third environment (e.g. if we were to try terraforming the recovery environment). Happy to chat more about this in a GMeet.

govwifi-frontend/security_groups.tf
```diff
@@ -173,5 +173,5 @@ resource "aws_security_group" "fe_radius_in" {
 
 data "aws_ip_ranges" "route53_healthcheck" {
   services = ["route53_healthchecks"]
-  regions  = var.is_production_aws_account ? ["eu-west-1", "us-east-1", "us-west-1"] : ["global"]
+  regions  = ["eu-west-1", "us-east-1", "us-west-1"]
```
camdesgov, line 176: Unfortunately we need to keep `global` here otherwise the health checks may not work in new environments. I agree this seems bewildering and counter intuitive, however it's what AWS recommends. AWS sent a detailed reason for this, which has been copy pasted in to the the comment made on Jul 23rd on this Trello card: https://trello.com/c/gXBNk1mP/1132-radius-route53-healthchecks-failing-due-to-incorrect-ip-range . There is also a shorter explanation here: https://github.com/alphagov/govwifi-terraform/commit/6dceb52b68749280af5ad7c96dbeb1ad3d38b53f. Happy to chat more over GMeet if it helps

cbaines: Ah, OK, I only went one commit deep in to the history, and didn't get to 6dceb52b68749280af5ad7c96dbeb1ad3d38b53f unfortunately.

I'm still a bit confused by this, but I think combining the lists should work, right? So specify those three specific regions plus global?

govwifi-frontend/certs.tf
cbaines: Discussed on Slack, I believe the consensus is to stick with the `is_production_aws_account` related behaviour in this Pull Request, and then try and move to specifying a `bucket_prefix` which doesn't include `rack_env` or the `env_subdomain`.

govwifi-frontend/security_groups.tf
cbaines: I've now gone ahead and switched to specifying all 4 IP ranges, let me know what you think?

camdesgov: Yeah I reckon this is fine

camdesgov (approved):

PR #642, opened 2021-12-06
Remove an old SNS subscription

cbaines:
### What
Remove an old SNS subscription

The subscription exists in production, but doesn't look to have a
valid endpoint. It doesn't exist in staging.

### Why
I believe there was a transition to a new topic/subscription beginning
back in 2018 [1].

1: 184998c48748e3eaa22842cfc34a7a5ff3dbfead

### DIFF
--- a/govwifi-emails/s3sns.tf
+++ b/govwifi-emails/s3sns.tf
@@ -189,16 +189,6 @@ EOF
 
 }
 
-# Subscription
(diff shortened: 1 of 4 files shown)


sarahseewhy (approved):
This looks fine to me. 

Bravo on the digging out the history of the resource and adding it to the description, that was helpful.

PR #644, opened 2021-12-07
Backup mysql rds refactoring

cbaines:
### What
Don't specify defaults, and use local rather than a variable.

### Why
Just neatening things up, less Terraform is better.

Link to Trello card: https://trello.com/c/ZdQrfyg7/1898-backupmysqlrds-refactoring-in-govwifi-terraform


sarahseewhy (approved):
LGTM

PR #647, opened 2021-12-08
Add https:// to the VOLUMETRICS_ENDPOINT for the Logging API

cbaines:
### What
Add https:// to the VOLUMETRICS_ENDPOINT for the Logging API

### Why
This not having https:// was a regression introduced in
2e68072df39f596175aff35fc0e54b8c0453f25c when this was set via
Terraform directly, rather than having the task use a value from a
secret.

Not including a protocol can work, but it seems like it's necessary in
this case to get the Elasticsearch client to use port 443 to talk to
the endpoint.


koetsier (approved):
lgtm!

PR #651, opened 2021-12-09
Tweak the frontend certs bucket

cbaines:
### What

 - Remove a redundant count from the bucket resource
 - Remove related duplicated IAM bits
 - Add a parameter that contains the name of the bucket

### Why
Cleanup, plus creating the parameters necessary to make the changes in https://github.com/alphagov/govwifi-build/pull/517 work.

### DIFF
--- a/govwifi-frontend/certs.tf
+++ b/govwifi-frontend/certs.tf
@@ -20,3 +20,10 @@ resource "aws_s3_bucket" "frontend_cert_bucket" {
     enabled = true
   }
 }
+
+resource "aws_ssm_parameter" "frontend_cert_bucket" {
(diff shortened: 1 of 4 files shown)


govwifi-frontend/certs.tf
```diff
@@ -20,3 +20,10 @@ resource "aws_s3_bucket" "frontend_cert_bucket" {
     enabled = true
   }
 }
+
+resource "aws_ssm_parameter" "frontend_cert_bucket" {
+  name        = "/from-terraform/frontend-certs-${var.aws_region_name}-bucket"
```
sarahseewhy: Thanks for explaining the naming, how about `govwifi-terraform`?

sarahseewhy (approved):

PR #652, opened 2021-12-10
Use the correct role for sync_s3_to_elasticsearch scheduled task

koetsier:
Use the correct role for sync_s3_to_elasticsearch scheduled task


sarahseewhy (approved):
LGTM.

We'll need to roll this out to `staging` and `production` since I had to revert the change in `staging` in order to fix the issue with the Grafana instance.

PR #662, opened 2022-01-10
Remove user management related Terraform

cbaines:
### What
Remove user management related Terraform.

### Why
As this is not being used.


sarahseewhy (approved):
Let's give this a whirl

PR #663, opened 2022-01-10
Use Secrets Manager secrets for Sentry DSNs

cbaines:
### What
Use Secrets Manager secrets for Sentry DSNs

### Why
I think this is neater, consolidating secrets in AWS rather than
keeping some in govwifi-build.

I'm looking at this as I'm interested in removing the dependency on
govwifi-build for running Terraform.

Deploying this will require manually creating the relevant secrets in
AWS.


sarahseewhy (approved):
LGTM

PR #664, opened 2022-01-10
Don't use a secret for the account id

cbaines:
### What
Don't use a secret for the account id

### Why
Terraform can just look this up.

### DIFF
--- a/govwifi/staging-dublin/data.tf
+++ b/govwifi/staging-dublin/data.tf
@@ -2,3 +2,5 @@
 data "aws_route53_zone" "main" {
   name = "${local.env_subdomain}.service.gov.uk."
 }
+
+data "aws_caller_identity" "current" {}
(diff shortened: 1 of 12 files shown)


sarahseewhy (approved):
Sensible refactor. Just to confirm, because it's not in the PR description, I'm guessing there's a follow-up task of removing the `account-id` secret from Secrets Manager as part of this clean up. 

I appreciate it goes without saying but since there isn't a checklist or Trello card I wanted to dot that i and cross that t.

cbaines:
> Sensible refactor. Just to confirm, because it's not in the PR description, I'm guessing there's a follow-up task of removing the `account-id` secret from Secrets Manager as part of this clean up.
> 
> I appreciate it goes without saying but since there isn't a checklist or Trello card I wanted to dot that i and cross that t.

It is indeed something that's not required when applying these changes, but it should still be done. I've added a bit to the commit message about it.

PR #667, opened 2022-01-24
Fixes to user-signup-api lockdown work

camdesgov:
### What
Lambda code correction, wrong url was being generated for production. Production ENV var is not “production” but “wifi”

Add missing variable “notify_ips" in production

### Why
Errors need to be fixed.

https://trello.com/c/QSO59546/1889-lockdown-requests-coming-into-the-user-signup


seb-szy (approved):
All good, this is a missing variable and updated Lambda we introduced to deploy the code.

PR #670, opened 2022-02-16
Allow to use trusted Administrator IPs in the CIDR format

seb-szy:
### What
We need to accept IP values formatted as the CIDR standard to accommodate a new range of IP used by VPN DR.

### Why
Current code was enforcing use of single IPs.


szd55gds (approved):
LGTM

PR #672, opened 2022-02-18
Send staging London logs to Cyber for ITHC

camdesgov:
### What
Send staging London logs to Cyber for ITHC

### Why
Cyber are performing some tests on our systems and would like our logs to be temporarily forwarded to them. This change is implemented following the guidance here:
https://github.com/alphagov/centralised-security-logging-service

Unfortunately the log names need to be hardcoded in the variable array as terraform cannot perform interpolation here.

Link to Trello card (if applicable): https://trello.com/c/MwGpCYMb/1964-send-data-from-govwifi-log-group-to-cybers-kinesis-stream

### DIFF
--- a/govwifi-admin/log-subscription-filters.tf
+++ b/govwifi-admin/log-subscription-filters.tf
@@ -1,0 +1,20 @@
+# Cyber has requested we add these
+
+variable "staging_log_group_names" {
+  type = list(string)
+  default = [
+    "staging-admin-log-group",
+    "/aws/rds/instance/wifi-admin-staging-db/audit",
+    "/aws/rds/instance/wifi-admin-staging-db/error",
+    "/aws/rds/instance/wifi-admin-staging-db/slowquery"
+  ]
+}
+
+resource "aws_cloudwatch_log_subscription_filter" "log_subscription" {
+  count = var.env_name == "staging" ? length(var.staging_log_group_names) : 0
+
+  name            = "log_subscription_${count.index}"
+  log_group_name  = element(var.staging_log_group_names, count.index)
+  filter_pattern  = ""
+  destination_arn = "arn:aws:logs:eu-west-2:${var.cyber_account_id}:destination:csls_cw_logs_destination_prod"
+}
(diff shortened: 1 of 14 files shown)


govwifi-admin/log-subscription-filters.tf
```diff
@@ -0,0 +1,20 @@
...
+
+variable "staging_log_group_names" {
+  type = list(string)
+  default = [
+    "staging-admin-log-group",
+    "/aws/rds/instance/wifi-admin-staging-db/audit",
+    "/aws/rds/instance/wifi-admin-staging-db/error",
+    "/aws/rds/instance/wifi-admin-staging-db/slowquery"
+  ]
+}
```
cbaines, line 11: This doesn't look to vary, so it should probably be a local rather than a variable. I think that would also resolve the issue with variable interpolation.

govwifi/staging-dublin/main.tf
```diff
@@ -119,6 +119,7 @@ module "backend" {
   grafana_ip            = var.grafana_ip
 
   db_storage_alarm_threshold = 19327342936
+  cyber_account_id           = ""
```
cbaines, line 122: I think it would be cleaner to give this variable a default, rather than explicitly passing in the empty string where it's unused.

cbaines (changes requested):

govwifi-admin/log-subscription-filters.tf
sarahseewhy: That's good to know, thank you. Given this is throw away code which will be cleaned up on Thursday I'm happy for it to be merged.

govwifi/staging-dublin/main.tf
sarahseewhy: Again, this is throw away code. It's good to know for future additions.

sarahseewhy (approved):

govwifi-admin/log-subscription-filters.tf
cbaines: If we're going to merge it in, I still think it's necessary to go through the usual code review process.

PR #677, opened 2022-03-15
Prepare for terraform region consolidation

cbaines:
### What
Make some small Terraform changes around EIPs.

### Why
This is in preparation for trying to consolidate the Terraform for a single environment, as it means that the IP addresses can be passed around as values in Terraform, rather than being provided to various modules as variables.

Link to Trello card: https://trello.com/c/HublU99C/2065-remove-necessity-to-specify-a-region-when-running-govwifi-terraform


camdesgov:
Ahhh right. Thanks for explaining. Approved.

PR #684, opened 2022-04-28
Switch from RACK_ENV to RAILS_ENV for the admin app

cbaines:
### What
Switch from RACK_ENV to RAILS_ENV for the admin app

### Why
RACK_ENV worked when starting the app with rails server, but doesn't
work when directly running the app with Puma. Instead, set the
RAILS_ENV environment variable which is a more usual thing to set.

Also, this isn't great because some of the app configuration seems to
be in the govwifi-admin, and depend on setting the right RAILS_ENV for
a particular deployment (e.g. staging for staging).


koetsier (approved):
looks good to me

PR #685, opened 2022-05-03
Fix the RAILS_ENV environment variable in the admin task

cbaines:
### What
Fix the RAILS_ENV environment variable in the admin task

### Why
It should not be hardcoded to staging.


koetsier (approved):
👍

PR #688, opened 2022-05-25
Switch govwifi-admin to connect to the primary users DB

cbaines:
### What
Switch govwifi-admin to connect to the primary users DB

### Why
As this will allow deleting users from the admin app, something which
currently happens but has to be done by hand.

In staging, I believe the admin app is currently connected to the
primary database.

Link to Trello card: https://trello.com/c/yw5VYGdD/2265-a-super-admin-can-delete-user-credentials


camdesgov (approved):
Looks fine to me, approved.

PR #690, opened 2022-06-17
Make container name the same as app name

camdesgov:
### What

Make container name consistent with the name we use for the ECS service,
cluster and task definition. In our ECS task definitions we currently declare a
container name that is different from the app, task and cluster name:
[govwifi-terraform/user-signup-api-cluster.tf at
13c45e4791ba82dfaba89cf68847eb2d720c4825 ·
alphagov/govwifi-terraform](https://github.com/alphagov/govwifi-terraform/blob/13c45e4791ba82dfaba89cf68847eb2d720c4825/govwifi-api/user-signup-api-cluster.tf#L201)

This name is also referenced in our application code, see here for an example:
https://github.com/alphagov/govwifi-user-signup-api/search?q=container_name

The name used is "user-signup" but elsewhere we use "user-signup-**api**" to refer
to related resources like the ECS service. This should be consistent -
"user-signup-api" should be used everywhere, and references to "user-signup"
should be deleted.

### Why

Using different values for the container name, service name and cluster
name is confusing and makes writing the pipeline code more complicated.

Link to Trello card: https://trello.com/c/vf9rqnRW/1501-make-user-signup-container-name-references-in-terraform-deploy-process-consistent


cbaines (approved):
Sounds OK to me

cbaines:
I'm guessing this change needs making in the `aws_ecs_task_definition.user_signup_api_task` right?

Searching for `user-signup"` through the Terraform suggests that the scheduled tasks might need updating too?

camdesgov:
> 

Good spot. I’ve made some more changes. Could you re-approve?

govwifi-emails/s3sns.tf
```diff
@@ -190,16 +190,16 @@ EOF
 }
 
 # SNS topic to notify the new user-signup API when an email arrives
-resource "aws_sns_topic" "user_signup_notifications" {
-  name         = "${var.env_name}-user-signup-notifications"
-  display_name = "${title(var.env_name)} user signup email notifications"
+resource "aws_sns_topic" "user_signup_api_notifications" {
```
cbaines, line 193: You don't need to change this, but it's also OK to do so.

govwifi-emails/s3sns.tf
```diff
@@ -190,16 +190,16 @@ EOF
...
+  name         = "${var.env_name}-user-signup-api-notifications"
+  display_name = "${title(var.env_name)} user signup api email notifications"
```
cbaines, line 195: I don't think these bits here are related to the container name.

They're fine to change, but it would be clearer to do it in a separate commit.

cbaines:
> Good spot. I’ve made some more changes. Could you re-approve?

I think there's maybe more to change with the scheduled tasks, this was what I found when searching for `user-signup"`:

```
File: govwifi-api/user-signup-scheduled-tasks.tf
31:16:      "name": "user-signup",
128:16:      "name": "user-signup",
167:16:      "name": "user-signup",
244:16:      "name": "user-signup",
343:16:      "name": "user-signup",

File: govwifi-api/user-signup-api-cluster.tf
105:16:      "name": "user-signup",
201:25:    container_name   = "user-signup"
```

govwifi-emails/s3sns.tf
camdesgov: Removed these from this PR. I agree it should be part of another PR/commit. I will roll the container name changes out separately first.

govwifi-emails/s3sns.tf
camdesgov: See comment below

camdesgov:
> Have now updated these.

PR #695, opened 2022-07-06
Start the transition to frontend using fargate and a load balancer

cbaines:
### What
The commit adda a whole new set of resources to run frontend (RADIUS)
as a load balanced service using ECS Fargate.

### Why
The intent here is to run this in addition to the current EC2 setup,
then migrate EIPs over to the load balancer. For testing prior to
this, a new set of EIPs is used, and these will be deregistered once
the migration is complete.

Link to Trello card: https://trello.com/c/5zf9ZXj5/2399-add-udp-network-loadbalancers-in-front-of-the-radius-servers-to-increase-resiliency

### DIFF
--- a/govwifi-frontend/cluster.tf
+++ b/govwifi-frontend/cluster.tf
@@ -9,6 +9,17 @@ resource "aws_ecs_cluster" "frontend_cluster" {
   }
 }
 
+resource "aws_ecs_cluster" "frontend_fargate" {
+  name = "frontend-fargate"
+
+  capacity_providers = ["FARGATE"]
+
+  setting {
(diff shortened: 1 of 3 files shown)


camdesgov:
Looks fine to me. Approved. Good idea to launch a separate cluster.

You have a small typo in the PR comment:

> The commit adda a whole new set of resources to run frontend (RADIUS)

I think "adda" should be "adds"

cbaines:
> > The commit adda a whole new set of resources to run frontend (RADIUS)
> 
> I think "adda" should be "adds"

Thanks, I've fixed this now.

PR #701, opened 2022-07-12
Remove the template provider from the lockfiles

cbaines:
### What
Remove the template provider from the lockfiles

### Why
This provider is no longer necessary, as all instances of it's use
have been removed and the state has been updated with "apply
-refresh-only".

Link to Trello card: https://trello.com/c/PpqGUlLO/2423-get-terraform-working-on-apple-silicone-architecture-macbooks


szd55gds (approved):
LGTM

PR #710, opened 2022-07-20
Fix the fargate frontend in the ireland regions

cbaines:
### What
Fix the fargate frontend in the ireland regions

### Why
This fixes issues with the tasks starting because of access to container images and S3 buckets.


cbaines:
This depends on https://github.com/alphagov/govwifi-terraform/pull/709

camdesgov (approved):
LGTM

PR #713, opened 2022-07-21
Fix various deprecations

cbaines:
### What
Fix various deprecations resulting from the AWS provider upgrade.

### Why
Keeping things up to date is good.


camdesgov (approved):
Approved pending test in Staging

PR #714, opened 2022-07-21
Disable the Fargate frontend cluster

cbaines:
### What
Disable the Fargate frontend cluster

### Why
As it's not working yet, and this avoids alarms about it not
working. This can be enabled again once the internal load balancers
for the authentication and logging APIs are setup.


szd55gds (approved):
Looks ok to me

PR #717, opened 2022-07-21
Configure prometheus for the fargate frontend

cbaines:
### What
Configure Prometheus for the fargate frontend.

### Why
This functionality is important to keep.


camdesgov (approved):
LGTM

PR #721, opened 2022-08-08
Fargate frontend fixes

cbaines:
### What
Fargate frontend fixes

### Why
These changes are sufficient to get the smoke tests passing.

### DIFF
--- a/govwifi-frontend/security_groups.tf
+++ b/govwifi-frontend/security_groups.tf
@@ -211,19 +211,19 @@ resource "aws_security_group" "load_balanced_frontend_service" {
   }
 
   ingress {
-    description = "RADIUS traffic from load balancer"
+    description = "RADIUS traffic"
     from_port   = 1812
     to_port     = 1812
     protocol    = "udp"
-    cidr_blocks = [for ip in aws_eip.test_radius_eips.*.public_ip : "${ip}/32"]
+    cidr_blocks = ["0.0.0.0/0"]
   }
 
   egress {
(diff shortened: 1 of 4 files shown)


govwifi-frontend/security_groups.tf
```diff
@@ -211,19 +211,19 @@ resource "aws_security_group" "load_balanced_frontend_service" {
   }
 
   ingress {
-    description = "RADIUS traffic from load balancer"
+    description = "RADIUS traffic"
     from_port   = 1812
     to_port     = 1812
     protocol    = "udp"
-    cidr_blocks = [for ip in aws_eip.test_radius_eips.*.public_ip : "${ip}/32"]
+    cidr_blocks = ["0.0.0.0/0"]
```
camdesgov, line 218: Just curious - is this because AWS sends the traffic coming from the load balancer out over the internet then back into the cluster?

camdesgov (approved):
Approved, but would just double check the security group thing with Seb, to see if there is any way we can make this less open.

govwifi-frontend/security_groups.tf
cbaines: Unsure. The previous config assumed that the traffic was coming from the EIPs associated with the load balancer. But they could be coming from some local IP, or maybe the original source IP.

Anyway, the tasks only have internal addresses, so this security group is only securing against other things in the VPC.

camdesgov: Ah right, thanks for the explanation

PR #723, opened 2022-08-11
Add codebuild job for safe-restarter

camdesgov:
### What
Add codebuild job for safe-restarter
Also change the ECR policy in tools into a data resource to make it easier to reuse.

### Why
We are migrating our pipelines from Concourse to AWS.

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-238

### DIFF
--- a/govwifi-deploy/codebuild-saferestarter.tf
+++ b/govwifi-deploy/codebuild-saferestarter.tf
@@ -1,0 +1,86 @@
+resource "aws_codebuild_project" "govwifi_codebuild_saferestarter" {
+  name           = "govwifi-build-safe-restarter"
+  description    = "This project builds the safe-restarter image and pushes it to ECR"
+  build_timeout  = "5"
+  service_role   = aws_iam_role.govwifi_codebuild.arn
+  encryption_key = aws_kms_key.codepipeline_key.arn
+
+  artifacts {
+    type = "NO_ARTIFACTS"
+  }
+
+  environment {
+    compute_type                = "BUILD_GENERAL1_SMALL"
+    image                       = "aws/codebuild/standard:5.0"
+    type                        = "LINUX_CONTAINER"
+    image_pull_credentials_type = "CODEBUILD"
+    privileged_mode             = true
+
+    environment_variable {
+      name  = "DOCKER_HUB_AUTHTOKEN_ENV"
+      value = data.aws_secretsmanager_secret_version.docker_hub_authtoken.secret_string
+    }
+
+    environment_variable {
+      name  = "DOCKER_HUB_USERNAME_ENV"
+      value = data.aws_secretsmanager_secret_version.docker_hub_username.secret_string
+    }
+
+    environment_variable {
+      name  = "AWS_ACCOUNT_ID"
+      value = local.aws_account_id
+    }
+
+    environment_variable {
+      name  = "STAGE"
+      value = "staging"
+    }
+
+    environment_variable {
+      name  = "REPOSITORY_URI"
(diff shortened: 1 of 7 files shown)


govwifi-deploy/ecr.tf
```diff
@@ -6,32 +6,38 @@ resource "aws_ecr_repository" "govwifi_ecr_repo" {
...
+#         "ecr:GetDownloadUrlForLayer",
+#         "ecr:InitiateLayerUpload",
+#         "ecr:PutImage",
+#         "ecr:UploadLayerPart"
+#       ]
+#     }
+#   ]
+# }
+# EOF
+# }
```
cbaines, line 39: Maybe remove this rather than commenting out?

govwifi-deploy/codebuild-saferestarter.tf
```diff
@@ -0,0 +1,86 @@
+resource "aws_codebuild_project" "govwifi_codebuild_saferestarter" {
```
cbaines, line 1: I'd be interested in why this is separate from the pipeline resource in codebuild.tf?

govwifi-deploy/ecr.tf
camdesgov: Commented section now removed

govwifi-deploy/codebuild-saferestarter.tf
cbaines: Ok, but the codebuild bit works the same right, it's the deploy part of the codepipeline that can be omitted for the safe restarter?

I'm not really a fan of this `for_each` `var.app_names` pattern, as it makes tweaking the behaviour for specific apps difficult, using modules would be preferable in my opinion. But, if you're interested in sticking with it, maybe it's worth having several lists of apps, like `deployed_app_names` and `build_app_names`. That would mean that the safe restarter could only be in `built_app_names`, as it doesn't need deploying. Does that make sense? I think that would allow duplicating bits of the Terraform to try and customise the behaviour for specific apps.

govwifi-deploy/ecr.tf
```diff
@@ -40,28 +17,41 @@ resource "aws_ecr_repository" "govwifi_frontend" {
...
-  ]
+resource "aws_ecr_repository" "safe_restarter_ecr" {
+  name = "govwifi/staging/safe-restarter"
+}
+
+resource "aws_ecr_repository_policy" "govwifi_ecr_saferestater_policy" {
+  for_each   = toset(var.app_names)
+  repository = aws_ecr_repository.safe_restarter_ecr.name
+  policy     = data.aws_iam_policy_document.govwifi_ecr_repo_policy.json
 }
```
cbaines, line 31: Why is this ECR policy being set for each app name?

govwifi-deploy/codebuild-saferestarter.tf
camdesgov: Yep makes sense, I like the idea about splitting stuff into several lists of apps.  Working on implementing this now, I'll push the update on Tuesday. Consider this PR paused until then.

govwifi-deploy/ecr.tf
camdesgov: Each app has an image repo associated with it, and staging (and later production) needs to be able pull images from our tools account. The policy allows this.

szd55gds (approved):
LGTM and paired with Camille on this work.

govwifi-deploy/codebuild-saferestarter.tf
camdesgov: I've now separated things into two lists.

PR #725, opened 2022-08-15
Fix Staging Ireland Prometheus connectivity

cbaines:
### What
Fix Staging Ireland Prometheus connectivity

### Why
Adjust the security groups so that you can SSH in to the Ireland
Prometheus in Staging, and so that Grafana can connect to it too.


seb-szy (approved):
LGTM

PR #727, opened 2022-08-18
Migrate one EIP to the frontend Network Load Balancer

cbaines:
### What
This commit sets Terraform up to migrate the 2nd EIP (eu-west-1b,
eu-west-2b) from the EC2 instances to the Network Load Balancer.

### Why
To start the switch to the Network Load Balancer.


cbaines:
Depends on https://github.com/alphagov/govwifi-terraform/pull/726

camdesgov (approved):

PR #728, opened 2022-08-22
Attach some temporary IPs to the NLBs again

cbaines:
### What
Attach some temporary IPs to the NLBs again

### Why
This addresses the temporary issue where the NLB is only routing
traffic to tasks in the subnets where it has a presence, and tasks in
other subnets were repeatedly starting and stopping.


seb-szy (approved):
Chris has clarified all the syntax, all good.

PR #731, opened 2022-08-23
Make API access from frontend EC2 instances with no EIP work

cbaines:
### What
Make API access from frontend EC2 instances with no EIP work

### Why
With the loss of the EIPs, the tasks running on the EC2 instances have
lost access to the authentication and logging APIs.

Since the EIPs have switched across to the Network Load Balancer, this
shouldn't cause an issue for regular GovWifi traffic to these
IPs. However, it seems that there's still some traffic to the tasks on
these instances.

This commit should restore API access for these tasks, just in case
that's useful for some reason.


cbaines:
I don't think this'll be needed.

PR #732, opened 2022-08-24
Change PagerDuty to trigger off the NLBs

cbaines:
### What
Change PagerDuty to trigger off the NLBs

### Why
This is part of the migration to Network Load Balancers for frontend.


cbaines:
This depends on https://github.com/alphagov/govwifi-terraform/pull/730

camdesgov (approved):

PR #733, opened 2022-08-24
Tag all resources by environment

cbaines:
### What
Tag all resources by environment

### Why
This can help to check what environment a resources corresponds to, as
it's not always obvious on the AWS website.

This is more important now that there are more resources where the
environment name (wifi, staging, ...) hasn't leaked in to the resource
name.


camdesgov (approved):
Small typo in the second line of this commit message: https://github.com/alphagov/govwifi-terraform/pull/733/commits/c1f2304f617388542cd2518926ed360e974e49c3 Looks like `resources` should be`resource`. Other than that looks fine.

PR #736, opened 2022-08-26
Database backup codebuild

szd55gds:
### What
Add the terraform to include AWS Codebuild Project for the Database Backup module.
Create the AWS Tools account ECR Repo for Database backup images.

### Why
This is part of our migration from Concourse to AWS Developer Tools.
This task is not deployed as such and so only requires a CodeBuild Project.
The reference module is in repo: govwifi-database-backup.

### Link to Jira card (if applicable): 
https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?assignee=5b2a20739ba6d02662e2fc0b&selectedIssue=GW-246

### DIFF
--- a/govwifi-deploy/ecr.tf
+++ b/govwifi-deploy/ecr.tf
@@ -29,6 +29,14 @@ resource "aws_ecr_repository_policy" "govwifi_ecr_saferestater_policy" {
   policy     = data.aws_iam_policy_document.govwifi_ecr_repo_policy.json
 }
 
+resource "aws_ecr_repository" "database_backup_ecr" {
+  name = "govwifi/staging/database-backup"
+}
+
+resource "aws_ecr_repository_policy" "govwifi_ecr_database_backup_policy" {
+  repository = aws_ecr_repository.database_backup_ecr.name
+  policy     = data.aws_iam_policy_document.govwifi_ecr_repo_policy.json
+}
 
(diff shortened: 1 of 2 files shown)


camdesgov (approved):

govwifi-deploy/ecr.tf
```diff
@@ -29,6 +29,14 @@ resource "aws_ecr_repository_policy" "govwifi_ecr_saferestater_policy" {
...
 }
 
+resource "aws_ecr_repository" "database_backup_ecr" {
+  name = "govwifi/staging/database-backup"
+}
+
+resource "aws_ecr_repository_policy" "govwifi_ecr_database_backup_policy" {
+  repository = aws_ecr_repository.database_backup_ecr.name
+  policy     = data.aws_iam_policy_document.govwifi_ecr_repo_policy.json
+}
```
cbaines, line 39: It would be better to use the `govwifi_ecr_repo` resource and `govwifi_ecr_repo_policy` resource instead, rather than using it for some apps, then creating separate resources in the Terraform for frontend, the safe restarter and the database backup app.

szd55gds: Thanks for your comment Chris. I refer you to the same point in the safe-restarter. Next quarter we will probably refactor the 'built apps' to make it match the pattern for the 'deployed apps' array.

PR #741, opened 2022-09-16
Update linter version

camdesgov:
### What
Update linter version

### Why
The linter was failing with the following error: "Failed to initialize plugins; TFLint is not compatible with this version of the "aws" plugin. A newer TFLint or plugin version may be required. Plugin version: 10, Client versions: [11]"

This change upgrades the linter, but also temporarily stops it from running in Github actions. This is because there are a large number of issues to fix before the linter will run successfully. Once these are fixed we will re-enable the linter.

Jira card: https://technologyprogramme.atlassian.net/browse/GW-504

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,13 +6,13 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: lint
-        run: |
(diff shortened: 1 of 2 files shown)


szd55gds (approved):
LGTM

PR #744, opened 2022-10-07
Update github actions. Fix validator, update linter

camdesgov:
### What
Update github actions. Fix validator, update linter

### Why
Brew is no longer installed by default on the ubuntu-20.04 container. We use brew in our github actions tests, and it's removal has caused these to fail. Adding the recommended workaround. More information is available here: https://github.com/actions/runner-images/issues/6283


szd55gds (approved):
LGTM

szd55gds (approved):
LGTM

PR #745, opened 2022-10-14
Add authentication-api to built apps list

szd55gds:
### What
Add authentication-api to built apps module list because we're ready to test it

### Why
We were waiting until we had solved the naming conflict authentication Vs authorisation. This has now been completed, settling on authentication.


camdesgov (approved):
lgtm

PR #747, opened 2022-10-19
Revert "Introduce codepipeline and codebuild modules"

camdesgov:
Reverts alphagov/govwifi-terraform#746

Merged some other changes in error. Reverting


szd55gds (approved):
Happy with this

PR #754, opened 2022-11-11
Add sns smoke-tests policy permissions

camdesgov:
### What
Add sns smoke-tests policy permissions
Disable scheduled smoke-tests

### Why
Disabling scheduled smoketests for now as AWS alerting isn't working


szd55gds (approved):
LGTM

PR #759, opened 2022-12-06
Ignore task definition changes after initial creation

camdesgov:
### What
Ignore task definition changes after initial creation

### Why
As we are now using AWS's native ECS tools to deploy, and they change the task definition version after every deploy, we will need to add an ignore statement to the code so that terrafrom will not try to continually roll back the task definition to an earlier version every time we apply it.


szd55gds (approved):
LGTM

PR #764, opened 2022-12-19
Change the names of codebuild projects to match a standard

szd55gds:
### What
Change the name of codebuild projects to match a standard.
Remove extraneous white space in some files.
Remove comments used during development.

### Why
It was slow to find a project in a list and not sustainable when we add say our new Development environment and another set of projects.
A naming standard permits sorting by name to group a project tasks together.

### DIFF
--- a/govwifi-deploy/codebuild-convert-image.tf
+++ b/govwifi-deploy/codebuild-convert-image.tf
@@ -3,7 +3,6 @@
 
 resource "aws_codebuild_project" "govwifi_codebuild_project_convert_image_format" {
   for_each      = toset(var.deployed_app_names)
-  # name          = "govwifi-codebuild-convert-image-format-${each.key}"
   name          = "${each.key}-convert-image-format"
   description   = "This job converts the ECR image into a format usable by the ECS deploy stage"
   build_timeout = "5"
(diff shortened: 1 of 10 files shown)


govwifi-deploy/codebuild-convert-image.tf
```diff
@@ -3,7 +3,8 @@
 
 resource "aws_codebuild_project" "govwifi_codebuild_project_convert_image_format" {
   for_each      = toset(var.deployed_app_names)
-  name          = "govwifi-codebuild-convert-image-format-${each.key}"
+  # name          = "govwifi-codebuild-convert-image-format-${each.key}"
```
sarahseewhy, line 6: It looks like we're keeping that commented out line `# name = "govwifi-codebuild-convert-image-format-${each.key}"` is that intentional?

sarahseewhy (approved):
LGTM besides that one comment about the commented out line.

govwifi-deploy/codebuild-convert-image.tf
szd55gds: Good spot @sarahseewhy. Removed and pushed.

PR #770, opened 2023-01-18
Add a test environment called alpaca

deanwilson:
This does not build a full environment, but it should build more than not having it


camdesgov:
Closing this as we are raising another PR in it's place

PR #773, opened 2023-01-24
Enable Scheduled AWS Smoke Tests In Production

camdesgov:
### What
Enable Scheduled AWS Smoke Tests In Production

### Why
This is part of the Concourse to AWS migration work. The new smoke tests have now been built and tested in production, and have Slack alerting in place, so scheduled smoke tests can now be enabled. The smoke tests in Concourse have been paused so they do not conflict.


szd55gds (approved):
LGTM

PR #783, opened 2023-03-30
Update AMI For Prometheus & Grafana

camdesgov:
### What
Update AMI For Prometheus & Grafana

### Why
AWS has removed the old one from it's catalogue and it is no longer available for use. This commit replaces the AMI with the closest alternative. There is work in the schedule to eventually replace these instances with ECS containers.

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-832


szd55gds (approved):
LGTM - discussed in detail.

PR #786, opened 2023-04-20
Fix Error Caused by Slack Alert URL Secret

camdesgov:
### What
Fix Error Caused by Slack Alert URL Secret

### Why
The slack-alert-url secret should only be created in production. This commit adds a conditional to implement that.


deanwilson (approved):
Looks good, merge when green.

PR #787, opened 2023-05-03
Add Automatic Snapshot Volume Backups for Prometheus and Grafana

camdesgov:
### What

Add Automatic Snapshot Volume Backups for Prometheus and Grafana. This enables the data for Prometheus, and the Grafana configuration to be recovered if the volumes or instances are deleted. The snapshots are created and managed using Amazon Data Lifecycle Manager, which incurs no extra cost (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/snapshot-lifecycle.html).

### Why

It is best practice to have a backup of our data

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-858

### DIFF
--- a/govwifi-grafana/dlm.tf
+++ b/govwifi-grafana/dlm.tf
@@ -1,0 +1,68 @@
+
+resource "aws_dlm_lifecycle_policy" "grafana_ebs_volume_backup" {
+  description        = "Grafana EBS DLM lifecycle policy"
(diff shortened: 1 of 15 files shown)


szd55gds (approved):
LGTM and much needed bit of tech debt, thank you

PR #792, opened 2023-05-26
Bring Dev/Alpaca environment in line with other environments

camdesgov:
### What
Bring Dev/Alpaca environment in line with other environments

### Why
The changes relating to the Prometheus volume backups had not been fully applied to the alpaca environment. This commit fixes that.

Prometheus volume backup work for reference:
https://github.com/alphagov/govwifi-terraform/pull/787


deanwilson (approved):
Merge away

PR #793, opened 2023-05-26
Tighten permissions on s3 buckets related to pipelines

camdesgov:
### What
All cross account roles now only have read only access to pipeline assets.

### Why
The previous permissions were insecure and theoretically allowed a staging role to edit a production asset.

**More details:** 
The native Codepipeline deployment stage ( see code here: https://github.com/alphagov/govwifi-terraform/blob/master/govwifi-deploy/codepipeline-admin.tf#L62-L82) requires access to the codepipeline bucket which lives in the govwifi-tools AWS account. There is not a way to reliably grant cross account roles access to specific objects in the bucket, whilst denying them access to others AND having a pipeline that will deploy to multiple environments...e.g. staging through to production.

If Codepipeline allowed us to set up directories in which you could place assets relating to each specific STAGE of the pipeline, then we could create staging assets in a staging directory, and production assets in a production directory, and so on. We could also restrict the cross account role to specific directories (e.g. govwifi-codepipeline-bucket/adminpipeline/staging). Unfortunately Codepipeline does not work this way.

The names of the directories in the asset bucket are autogenerated, and any assets are lumped into a randomly named directory, there is no way of distinguishing between staging and production assets that are part of the same pipeline. Hence giving all the cross account roles read only access is the next best solution. The assets themselves do not contain any production credentials or secrets, so there is no risk of credential breaching if they are accessed by other accounts.

**Link to Jira card (if applicable):** 
https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-859

### DIFF
--- a/govwifi-api/iam.tf
+++ b/govwifi-api/iam.tf
@@ -216,7 +216,8 @@ resource "aws_iam_policy" "crossaccount_tools" {
         {
             "Effect": "Allow",
             "Action": [
-                "s3:*"
+                "s3:GetObject",
+                "s3:ListBucket"
             ],
             "Resource": [
                 "arn:aws:s3:::govwifi-codepipeline-bucket",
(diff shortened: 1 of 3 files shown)


deanwilson (approved):
This solves the underlying issue in a non-intrusive way. Looks good - merge away.

PR #795, opened 2023-05-31
Remove radius IPS from secrets for smoke tests.

koetsier:
### What
Remove radius IPS from secrets for smoke tests.
### Why
The radius IPs are already in the Govwifi-build repo and replicating them in secrets can only introduce errors.
It Staging and development, they can be derived from the frontend module. At a later data we can hopefully remove them as constants altogether.

### DIFF
--- a/govwifi-smoke-tests/codebuild.tf
+++ b/govwifi-smoke-tests/codebuild.tf
@@ -74,7 +74,7 @@ resource "aws_codebuild_project" "smoke_tests" {
 
     environment_variable {
       name  = "RADIUS_IPS"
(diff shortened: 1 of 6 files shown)


camdesgov:
Heya, looks we're unwittingly working on the same issue. This PR automatically generates the IPs as secrets when they are created: https://github.com/alphagov/govwifi-terraform/pull/796 I think it's better to have them managed dynamically as secrets rather than as statically updated values in govwifi-build..... Shall we have a chat about this on Tuesday?

camdesgov:
Shall we close this PR, based on the meeting this morning?

PR #797, opened 2023-06-06
Move SMOKE_TEST_IPS variable to scheduled task

koetsier:
The session deletion scheduled task uses this, not the actual logging app


deanwilson (approved):
Merge away.

PR #798, opened 2023-06-16
Autogenerate database passwords on first terraform run.

camdesgov:
### What
This new code:

- First sets the database username and password
- Database username and password is then passed to RDS instance on initial creation
- When the RDS instance has been created for the corresponding app, the secret is then updated with the host address etc
- ignore_changes and depends_on are used to ensure correct order of creation and that existing secrets are not overwritten

Releasing to production.  After this code is merged the following secrets need to be imported into terraform immediately (terraform should then report no changes).
- rds/admin-db/credentials
- rds/users-db/credentials
- rds/session-db/credentials

### Why
Entering the secrets manually is time consuming and unnecessary. Having them automatically generated will make it easier and quicker to bring up new GovWifi environments on demand. This will help with our DR (Disaster Recovery) time.

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-905


szd55gds (approved):
LGTM

PR #799, opened 2023-07-06
Increase timeout of smoke tests to 10 minutes

koetsier:
When attempting to connect to Notify, the smoke tests have a 5 minute timeout set in the code. The job is also set to time out in 5 minutes, which means that the job is usually terminated before the code can time out and print a helpful message making debugging easier

This commit increases the timeout of the AWS task to 10 minutes so that the timeout occurs in the code.

The jobs are run every 15 minutes, so they should still run sequentially.


camdesgov (approved):
Looks good

PR #801, opened 2023-07-12
    Give FIDUS Access To Our EC2 Instances

camdesgov:
### What
Give FIDUS Access To Our EC2 Instances

### Why
They are performing the annual IT Healthcheck

PR #802, opened 2023-07-18
Revert "    Give FIDUS Access To Our EC2 Instances"

camdesgov:
Reverts alphagov/govwifi-terraform#801

FIDUS have now finished the IT Healthcheck. We can revoke their access.


szd55gds (approved):
LGTM

PR #804, opened 2023-07-20
Correction to randomly generated password length

camdesgov:
### What
Correction to randomly generated password length

### Why
The MySQL password limit is 32 characters. Reference: https://dev.mysql.com/doc/refman/8.0/en/assigning-passwords.html


szd55gds (approved):
LGTM merge when ready

PR #807, opened 2023-07-28
Change AMI To Data Reference

camdesgov:
### What
Change AMI To Data Reference

### Why
AMI id's are regional, but the AMI names are global. Changing this to a data reference and using an AMI name, avoids terraform errors when we deploy our code to the eu-west-1 region

Link to Jira card (if applicable): https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-1012


seb-szy (approved):
LGTM

PR #810, opened 2023-08-11
Remove Redundant IAM Users

camdesgov:
### What
These IAM users are not needed.
- govwifi-pipeline-deploy-prod
- govwifi-pipeline-deploy-smoketest Removing them and related policies as part of best practice.

### Why
The IAM user _govwifi-pipeline-deploy-smoketest_ has not been accessed for 204 days 
The IAM user _govwifi-pipeline-deploy-prod_ has never been accessed.

These users appear to be left over from when GovWifi was using Concourse for it's deployment and are no longer needed (the new codepipeline deployment uses the assume role facility to gain permissions, rather than accessing resources through an IAM user, so these old users are therefore redundant).


szd55gds (approved):
LGTM

PR #814, opened 2023-08-21
GW-1075 ec2 reboot

leonaAtkins:
### What
Ensure  EC2 Instances restart on a regular schedule so security updates can always be applied

### Why
Some security updates can only be fully applied after a restart. We want to automate this process so that we will not have to endure the manual toil.

[Jira Issue Ticket GW-1075](https://technologyprogramme.atlassian.net/browse/GW-1075)


szd55gds (approved):
LGTM, walked through in detail, merge when ready.

PR #822, opened 2023-09-27
Increase the timeout of smoke-tests

camdesgov:
### What
Increase the timeout of smoke-tests

### Why
We experiencing quite a few false flags, because we are waiting for Notify to respond.


koetsier (approved):
👍

PR #827, opened 2023-10-11
Correction to database engine version

camdesgov:
### What
Correction to database engine version

### Why
Database engine version needs to be set to 8.0


szd55gds (approved):
LGTM merge when ready

PR #833, opened 2023-11-03
Add GovWifi Docs DNS Records To Terraform

camdesgov:
### What

These two DNS Records exist in our production AWS account: https://dev-docs.wifi.service.gov.uk/
https://docs.wifi.service.gov.uk/
However were are not in our terraform. Now added in new module.

DNS record for the product page docs have been added to the new module, and moved from the admin module as this is more logical.

### Why

All our DNS records should be in terraform, so we can recreate our infrastructure 

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?assignee=5e45277d29e6d00c970616c6&selectedIssue=GW-1263

### DIFF
--- a/govwifi-admin/route53.tf
+++ b/govwifi-admin/route53.tf
@@ -9,11 +9,3 @@ resource "aws_route53_record" "admin" {
     evaluate_target_health = true
   }
 }
-
-resource "aws_route53_record" "www" {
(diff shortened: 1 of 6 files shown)


szd55gds (approved):
LGTM merge when ready

PR #834, opened 2023-11-07
fix: [GW-1180] add ignore tags and bump tf version

leonaAtkins:
### What
When performing a plan, terraform would always show a change for 'tags' which should have been already set, this happens because this tags are computed in the default_tags of the provider, which isn't supported, so terraform computes these every time.

### Why
As tags are computed, terraform always thinks have checked, [re: issue](https://github.com/hashicorp/terraform-provider-aws/issues/18311#issuecomment-1544330448)
A related bug to this was fixed in version 1.4.6, [again described in this github issue](https://github.com/hashicorp/terraform-provider-aws/issues/18311#issuecomment-1544330448)

Note: this PR pushes the TF version from 1.1.8 to 1.6.3 I've run this against Alpaca and Staging with no side effects.

[GW-1180](https://technologyprogramme.atlassian.net/browse/GW-1180)

[GW-1180]: https://technologyprogramme.atlassian.net/browse/GW-1180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


szd55gds (approved):
LGTM, I ran terraform against production with zero significant changes. Merge when ready :-)

PR #838, opened 2023-11-28
Update DNS Records For Documentation Pages

camdesgov:
### What
Update DNS Records For Documentation Pages

### Why
Remove connection to PaaS

Add www counterpart records so our domain names will work with Github pages. See here for more information:
https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/about-custom-domains-and-github-pages


leonaAtkins (approved):
LGTM :)

PR #840, opened 2023-12-01
fix: [GW-1006] lock down health check group.

leonaAtkins:
### What
Create health check group to hold aws ec2 ip's, there is a 60 rule limit on SG's so had to create a map to hold all the possible IP's for dublin.

### Why
Port 3000 was open to public traffic, as it's only used for the health check reporting this can be locked down to AWS ip's, this was highlighted in the Security Check, rest need to be publicly accessible. 

[GW-1006](https://technologyprogramme.atlassian.net/browse/GW-1006)


szd55gds (approved):
Looks ok to me, **merge next week please Leona**, we don't deploy Friday pm and we're very close to that.

szd55gds (approved):
Yup, this looks ok, merge when ready

PR #852, opened 2024-01-29
docs: fix broken links.

leonaAtkins:
### What
Some broken links in the readme.

### Why
Old links to PaaS left over from migration needed to the updated so we can find our docs.

Part of BCP/DR exercise


szd55gds (approved):
LGTM merge when ready

PR #859, opened 2024-03-13
bug: [GW-278] put correct BE VPC CDIR into local Variable for Dublin

leonaAtkins:
### What
Correct VPC ID and store them in locals so they are in only 1 place.

### Why
Bug with previous deployment where wrong VPC id was used.

[GW-278](https://technologyprogramme.atlassian.net/browse/GW-278)

[GW-278]: https://technologyprogramme.atlassian.net/browse/GW-278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,10 +2,10 @@ name: govwifi-terraform-linting
 on: [push, pull_request]
 jobs:
   lint:
-    runs-on: macos-latest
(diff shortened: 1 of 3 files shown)


szd55gds (approved):
LGTM merge when ready

PR #867, opened 2024-05-03
Fix bastion user-data issue #GW1424

szd55gds:
### What
Bastions will not rebuild due to AWS changing the method to install cloudwatch.
Also noticed, some legacy code removed.
Changes to IAM & SGs to permit cloudwatch to talk to the API
Updated AMI
Workaround for sudo cat > [file]

### Why
We were unable to build / rebuild bastions because user-data would not execute.
This prevented access to the instances because ssh keys were not inserted.
We need to monitor these systems and log groups were not created and updated.
Existing AMI did not function fully.

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?assignee=5b2a20739ba6d02662e2fc0b&selectedIssue=GW-1424

### Testing
Remove cloudwatch log groups and streams.
Terminate the bastion instance in an AWS account.
Deploy new instance (which will create new log groups / streams) via terraform


govwifi/alpaca/london.tf
```diff
@@ -57,7 +57,7 @@ module "london_backend" {
   frontend_radius_ips = local.frontend_radius_ips
 
   bastion_instance_type     = "t2.micro"
-  bastion_ssh_key_name      = "govwifi-alpaca-bastion-20230120"
+  bastion_ssh_key_name      = "alpaca-bastion-20230120"
```
leonaAtkins, line 60: maybe this should be a local variable so we're not repeating it.

szd55gds: maybe but i'm not going to add it this time, I have to revisit so will keep this PR clean and addressing the jira ticket. Some other tasks to do on other tickets e.g. AMI :-)

leonaAtkins (approved):
LGTM

PR #872, opened 2024-05-29
Export certificates

koetsier:
### What
Enable the Admin portal to export Certificates to an S3 bucket
### Why
So that the FreeRadius server can pick up these certificates during the safe restart and start accepting client certificates

https://technologyprogramme.atlassian.net/browse/GW-1683


govwifi-admin/cluster.tf
```diff
@@ -113,6 +113,12 @@ resource "aws_ecs_task_definition" "admin_task" {
         },{
           "name": "ELASTICSEARCH_ENDPOINT",
           "value": "https://${var.elasticsearch_endpoint}"
+        },{
+          "name": "S3_CERTIFICATES_BUCKET",
+          "value": "${var.frontend_cert_bucket}"
+        },{
+          "name": "S3_CERTIFICATES_OBJECT_KEY",
+          "value": "eap_tls_certificates.zip"
```
leonaAtkins, line 121: maybe the filename could be a variable.

leonaAtkins (approved):

govwifi-admin/cluster.tf
koetsier: Actually I decided to do your suggestion anyway. So we can share the S3 bucket/key with frontend and the admin portal

leonaAtkins:
@koetsier just checking, do the certs only exist in London, I don't see a Dublin config and just checking this is expected, does this get copied to Dublin in another way? replication?

koetsier:
> @koetsier just checking, do the certs only exist in London, I don't see a Dublin config and just checking this is expected, does this get copied to Dublin in another way? replication?

Yes, that gets replicated

leonaAtkins (approved):

PR #876, opened 2024-06-14
Remove Unused ECR Repos

camdesgov:
### What

Remove Unused ECR Repos

### Why

The ECS tasks were updated to reference  the ECR images in our tools account (where the GovWifi pipelines live) in late 2022, during the migration from Concourse and the introduction of AWS Codepipeline.

Prior to this, the ECR images were stored in their respective AWS environments (e.g. development, staging, production). These were not removed when Concourse was retired and Codepipeline went live as we wanted to be absolutely sure nothing was using them. Now that enough time has elapsed on the new system the ECR images in Development/Staging/Production should be deleted to reduce confusion and costs.

https://technologyprogramme.atlassian.net/browse/GW-1177

### DIFF
--- a/govwifi-admin/cluster.tf
+++ b/govwifi-admin/cluster.tf
@@ -8,11 +8,6 @@ resource "aws_cloudwatch_log_group" "admin_log_group" {
   retention_in_days = 90
 }
 
-resource "aws_ecr_repository" "govwifi_admin_ecr" {
-  count = var.ecr_repository_count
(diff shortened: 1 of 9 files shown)


szd55gds (approved):
LGTM, merge when ready

leonaAtkins (approved):
Interesting you needed to update the providers on Staging and London, what provers where missing?

PR #884, opened 2024-07-31
GW-1827 add the DEPLOY_ENV to terraform

iram-shehzadi:
### What
Add the DEPLOY_ENV to terraform for the authentication api

### Why
Describe why the change was necessary

[GW-1827](https://technologyprogramme.atlassian.net/browse/GW-1827)

[GW-1827]: https://technologyprogramme.atlassian.net/browse/GW-1827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/govwifi-api/authentication-api-cluster.tf
+++ b/govwifi-api/authentication-api-cluster.tf
@@ -43,7 +43,7 @@ resource "aws_ecs_task_definition" "authentication_api_task" {
           "name": "DB_HOSTNAME",
           "value": "${var.user_rr_hostname}"
         },{
-          "name": "RACK_ENV",
+          "name": "DEPLOY_ENV",
           "value": "${var.rack_env}"
         },{
           "name": "SENTRY_CURRENT_ENV",


govwifi-api/authentication-api-cluster.tf
```diff
@@ -43,7 +43,7 @@ resource "aws_ecs_task_definition" "authentication_api_task" {
           "name": "DB_HOSTNAME",
           "value": "${var.user_rr_hostname}"
         },{
-          "name": "RACK_ENV",
+          "name": "DEPLOY_ENV",
           "value": "${var.rack_env}"
```
leonaAtkins, line 47: So does this now mean it relates to the environment it's deployed to, ie (alpaca, staging, prod) rather than the internal Ruby environment, if so should we change var.rack_env to var.env?

koetsier:
this is no longer needed because we don't use DEPLOY_ENV anymore

PR #913, opened 2024-11-20
Enhance bootstrap script

szd55gds:
### What
Added significant functionality to this script

- uncomments 2 x secrets-manager.tf file code blocks
- creates intitial DNS entry and outputs nameservers for production incorporation to file: ns-records.json
- manipulates ns-records.json to insert real ns records into Production domain wifi.service.gov.uk
- creates access logs' buckets
- creates state bucket
- initialise terraform

### Why
We want to automate and script stuff so that it's lightning fast and repeatable

[GW-xxx](https://technologyprogramme.atlassian.net/browse/GW-xxx)

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -283,7 +283,7 @@ EOF
 # Use jq to extract NS records
 NS_RECORDS=$(echo "$JSON_INPUT" | jq -r '.DelegationSet.NameServers[]')
 
-# Create the JSON file
+# Create the JSON payload
 
 pwd
 cat <<EOF > ns-records.json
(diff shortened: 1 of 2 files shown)


scripts/bootstrap.sh
```diff
@@ -14,20 +14,30 @@
 #   modify the Makefile to add the new environment and preserve the previous version temporarily
 #   generate a key pair and store in govwifi-build repo secrets
 #   insert public key and key names into terraform files for the new environment
+#   uncomments 2 x secrets-manager.tf file code blocks
+#   creates intitial DNS entry and outputs nameservers for production incorporation to file: ns-records.json
+#   manipulates ns-records.json to insert real ns records into Production domain wifi.service.gov.uk
+#   creates access logs' buckets
+#   creates state bucket
+#   initialise terraform
+
```
leonaAtkins, line 23: Could add
```
# This script 'might' fail on a Windows/Linux system due to the syntax of sed, if this happens, change ` -i '' ` to ` -i ` from the sed command.
```

PR #945, opened 2025-05-14
Run A Lighter Version Of Smoketests In Production

camdesgov:
### What
Run A Lighter Version Of Smoketests In Production

### Why
The automated SMS journey routinely fails, due to issues with notify, this is not a reflection of system failing but an issue with the API call. This is giving false warnings about our production systems so this part of the test has been disabled in the production environment, but will still be used in staging and when we deploy new code.
https://github.com/GovWifi/govwifi-smoke-tests-light


leonaAtkins (approved):
Looks like, but just a suggestion you could have had the repo change dependant upon the enviroment eg if staging then smoke-tests else smoke-test-light etc.

PR #949, opened 2025-06-12
Update EC2 AMIs (Grafana, prometheus, bastion)

camdesgov:
### What
Update EC2 AMIs

### Why
The old AMIs can no longer be located on AWS. This caused an error when updating terraform. Updated path for AWS AMIs to new ones. Also added extra logging to user-data and fixed bash statement at beginning of the user-data file that was causing an error.

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?assignee=5e45277d29e6d00c970616c6&selectedIssue=GW-2302


koetsier (approved):
awesome 😁

PR #1025, opened 2026-06-03
feat: [GW-2828] initial work for moving smoke tests alerts to chatbot

leonaAtkins:
### What
Move smoke tests alerts to use the chatbot configuration

### Why
To simply and reuse existing configuration, reduce security attack vector by removing a lambda dependency with unsupported Node version.

[GW-2828](https://technologyprogramme.atlassian.net/browse/GW-2828)


ethancane (approved):
Approving this without insight so you can move along with your work.

Hope this helps!

ethancane (approved):

iram-shehzadi (approved):

PR #1031, opened 2026-06-26
fix: [GW-2380] fix format of UUID

leonaAtkins:
### What
Output the UUID not the JSON :)

### Why
so that cyber can read the uuid correctly.

[GW-2380](https://technologyprogramme.atlassian.net/browse/GW-2380)


ethancane-digital-co (approved):
LGTM

PR #1042, opened 2026-08-03
Adjust trigger metrics.

leonaAtkins:
### What
Adjust alarm metrics for the Radius can't connect to api

### Why
It's too sensitive, triggers when the API Scales up, when a new task  is created, Radius doesn't give this new task any grace period (needs to be fixed), so it tries to connect to it straight away, finds it can't then errors, there is no actual error, it's not just waiting long enough for the task to become stable! 
I've raised the threshold because 10 is way to low, when this things kicks off there are 100's of errors in the logs, also we need to give time for the task to become stable, so 5 minutes, in a 10 minute period should cover this.
For a 'real' outage, it will trigger after 5 points have be exceeded in a 10 minute period, other alarms would have kicked off by this point.

N/A no jira ticket, routine maintenance.


oisinmulvihill (approved):
This looks good to me!

🆗🚢

